PR #69, opened 2022-12-15
Slither Printer Integration

### LINKED ISSUE #52: Integrate slither's echidna printer
As @0xalpharush and @ggrieco-tob mentioned, we'll want to integrate slither's echidna printer to mine more interesting values and more importantly, contextualize them, for our `ValueSet`.

We currently mine some values from AST ourselves, but it's not perfect. We handle hex, decimal numbers, but possibly not numbers like `1 ETH` or `1e8`: https://github.com/trailofbits/medusa/blob/8036697794481b7bf9fa78c922ec7fa6a8a3005c/fuzzing/valuegeneration/value_set_from_ast.go#L13-L41

Reference for the echidna printer:
https://github.com/crytic/slither/blob/b96beeaa5742d5d7a862e964bc72b3cab67623f4/slither/printers/guidance/echidna.py#L350

anishnaik:
This PR resolves #52 and integrates slither's echidna printer for medusa. 

**How it works:**
1. The project is compiled
2. Slither is run on the _artifacts_ that were generated during compilation. If this fails, that's it, we do not try to re-compile. Thus, we are running with the `--ignore-compile` flag. This will cause issues with repos that use multiple frameworks for compilation.
3. The output is captured as a JSON object and unmarshaled into the `SlitherData` struct.
4. Currently, we only use the `constants_used` object and map it into a `Constants` object that is query-able and easier to use from a golang perspective. 
5. All constants from `Constants` are added to the base value set of the fuzzer.


anishnaik:
Closing since #530 was merged.

PR #75, opened 2022-12-27
Add support for optimization mode

### LINKED ISSUE #71: Add support for `optimization` mode
Optimization mode can be treated as another **test case provider**, similar to `property` or `assertion` mode. This is a great issue for someone who is starting to work with medusa because it touches on medusa's most critical packages, how medusa uses the subscribe/publish model to handle fuzz testing, and how easy it is to create a new testing strategy in medusa.

I would recommend understanding how Echidna's optimization mode works first. At a high level, the optimization mode in Echidna tries to maximize some value `x`. It _probably_ does this by storing the latest maximum, `max`,  and upon the running of a new call sequence will check whether the new max is greater than `max`. If so, it overwrites `max` with the new maximum. Realistically, it doesn't matter what `x` is, what you care about is the return value of the `echidna_opt` (`medusa_opt`) function.

This can be done in medusa as well, please refer to `test_case.go`, `test_case_assertion.go`, `test_case_assertion_provider.go`, `test_case_property.go`, and `test_case_property_provider.go`. Similarly, you will probably create a `test_case_optimization.go` and a `test_case_optimization_provider.go`. 

Here are some key considerations:
1. Support for optimization mode in the `config` and CLI. Should be disabled by default.
2. Use the property and assertion test case providers as examples on how to build out optimization mode.
3. Create a unit test or two that runs medusa in optimization mode.

tarunbhm:
closes #71 

- [x] unit test cases


anishnaik:
@tarunbhm do you want to go ahead and update this PR with the latest master branch? Happy to get this merged ASAP. Apologies for the delay.

ggrieco-tob:
I'm taking care of this. It seems that it should be ready for another review. I added some code to actually show the sequence of transactions that maximized certain function (otherwise, it was not useful, since these tests should never fail)

anishnaik:
Hey @tarunbhm + @ggrieco-tob so I did a review of this PR and I added the capability to shrink a call sequence that optimized the test case's value. This was something that was missing from the PR. However, during this development I realized that we will have to put this PR on hold for now b/c of the inefficiency of the call sequence shrinking. The reason it is inefficient is because every time the value is maximized, we send a shrink request. however, after the shrinking is done, we do not mark the test case as failed b/c the next call or call sequence could in fact maximize the value even more. Then, another computationally intensive shrink request can be triggered on the next call in the call sequence. Thus, optimization mode doesn't really have a "stop" - it keeps going as long as the fuzzer keeps going. Since we cannot do a one-time shrink request at the moment, this PR will have to be on hold.

Ideally, we could call `FuzzerWorker.shrinkCallSequence` but that does not work given the way medusa is structured and how the fuzzing loop is set up. We will have to wait until we integrate a higher-level testing API so that we can support _arbitrary shrinking requests_. This would allow us to only shrink optimization call sequences one time instead of a gajillion times.

anishnaik (approved):

Issue #66, opened 2022-12-14
Allow auto-complete filenames in sh/bash


ggrieco-tob:
Medusa requires a little bit more command line parameters than Echidna. This is fine, but there is one small detail that can be improved: auto-complete support (which echidna has already, in some degree). For instance, after typing `medusa fuzz --target` trying to autocomplete filenames will fail (will not give any suggestion at all). It will be nice to allow auto-complete here to speed up the typing.

PR #79, opened 2023-01-09
Autocompletion

ahpaleus:
This Pull Request is connected to the #66.

It enables autocompletion in the medusa tool and is available for bash, zsh, and PowerShell.
When a user uses a specific flag, it disappears from the suggestions.

image

just run:
```sh
$ ./medusa completion --help

Generate the autocompletion script for medusa for the specific shell.

Bash:
To load completions in the current shell session:

    source <(medusa completion bash)

To load completions for every new session, execute once:
- Linux:
      medusa completion bash > /etc/bash_completion.d/medusa

- macOS:
      medusa completion bash > /usr/local/etc/bash_completion.d/medusa

Zsh:
To load completions in the current shell session:

    source <(medusa completion zsh)

To load completions for every new session, execute once:

    medusa completion zsh > "${fpath[1]}/_medusa"

PowerShell:
To load completions in the current shell session:
PS> medusa completion powershell | Out-String | Invoke-Expression

To load completions for every new session, run:
PS> medusa completion powershell > medusa.ps1
and source this file from your PowerShell profile.

Usage:
  medusa completion [flags]

Flags:
  -h, --help   help for completion
```

follow instructions, and use `<tab><tab>` for better user experience.


Xenomega:
Before this is merged, I wanted to explore if there's a way to do autocomplete for file paths at least, even if you don't generate all of this and run these scripts for autocompletion of CLI commands. Most users presumably won't use this flow but will still want autocompletion for filenames, etc.

anishnaik:
@Xenomega autocompletion of file path is still satisfied whether or not you use the completion script. Just tested.

cmd/completion.go
```diff
@@ -0,0 +1,103 @@
...
+// runCompletionZsh generates and prints ZSH completion code to the STDOUT
+func runCompletionZsh(cmd *cobra.Command) error {
+	err := cmd.Root().GenZshCompletion(os.Stdout)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// runCompletionPowerShell generates and prints ZSH completion code to the STDOUT
```
anishnaik, line 96: update comment to say "...prints PowerShell completion..."

cmd/completion.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+		case "powershell":
+			err := runCompletionPowerShell(cmd)
+			if err != nil {
+				log.Fatalln("Error: unable to generate a powershell completion")
+			}
+		}
+	},
+}
+
```
anishnaik, line 73: Missing function comment here

cmd/completion.go
```diff
@@ -0,0 +1,103 @@
...
+PS> medusa completion powershell > medusa.ps1
+and source this file from your PowerShell profile.
+`
+
+// completionCmd represents the completion command
+var completionCmd = &cobra.Command{
+	Use:   "completion",
+	Short: "generate the autocompletion script for medusa for the specific shell",
+	Long:  generalComDesc,
+	Run: func(cmd *cobra.Command, args []string) {
```
anishnaik, line 49: I would recommend using the `RunE` parameter instead of `Run`. `RunE` allows you to return an error so instead of printing / log.Fatal you can just return an error.

cmd/completion.go
```diff
@@ -0,0 +1,103 @@
...
+		if len(args) == 0 {
+			fmt.Printf("Error: No shell specified")
```
anishnaik, line 51: Traditional golang convention is to make error messages all lowercase. Additionally, the "Error:" can be removed since that will be added automatically

So the error can be here (assuming you use `RunE`)
`return fmt.Errorf("no shell specified")`

Xenomega:
Ah disregard my previous comment then, maybe it's just that I remembered this wrong because I used cmd instead of powershell, or Windows didn't support this well at the time 🤷 . I thought for some reason this didn't work at the time.

cmd/init.go
```diff
@@ -9,15 +9,47 @@ import (
...
+				// none of the arguments are positional.
+				unusedFlags = append(unusedFlags, "--"+flag.Name)
+			}
+		})
+
+		// If a default platform is not specified, add a list of available platforms to the list of unused flags.
+		// If the --target or --out flag is set, it is assumed that the default platform is being used.
+		if len(args) == 0 && !(cmd.Flag("target").Changed || cmd.Flag("out").Changed) {
+			unusedFlags = append(unusedFlags, supportedPlatforms...)
+		}
```
anishnaik, line 47: So although I think this is a great idea I feel like it might be a bit confusing to the user? especially b/c specifying a platform is optional with `init`. Idk I'm fine with keeping it in but when I was testing it it felt a bit off.

@Xenomega what do you think?

cmd/fuzz.go
```diff
@@ -18,6 +19,26 @@ var fuzzCmd = &cobra.Command{
 	Long:  `Starts a fuzzing campaign`,
 	Args:  cmdValidateFuzzArgs,
 	RunE:  cmdRunFuzz,
+	// Run dynamic completion of nouns
+	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
anishnaik, line 23: The last concern I had was around the `--help` flag? Like once you add any flag to the command maybe the `--help` flag shouldn't show in the `unusedFlags` object? idk this is minor but wanted to get your guys thoughts

cmd/init.go
anishnaik: Or maybe we keep it and potentially generalize `!(cmd.Flag("target").Changed || cmd.Flag("out").Changed)`. Is it possible to make it such that "if any flag is used, then don't add supportedPlatforms to the list".

anishnaik: Maybe when we `VisitAll` we can maintain a boolean that says `flagUsed`. If `flag.Changed` is true for any flag, that `flagUsed` can be `true` and then we can check `if len(args) == 0 && !flagUsed { // add supportedPlatforms }`

anishnaik: [906a4ac](https://github.com/crytic/medusa/pull/79/commits/906a4aca25259f4bb03c93a23e14a216e7e5d272)

cmd/completion.go
anishnaik: Resolved in [906a4ac](https://github.com/crytic/medusa/pull/79/commits/906a4aca25259f4bb03c93a23e14a216e7e5d272)

anishnaik: Resolved in [906a4ac](https://github.com/crytic/medusa/pull/79/commits/906a4aca25259f4bb03c93a23e14a216e7e5d272)

cmd/completion.go
anishnaik: Resolved in [906a4ac](https://github.com/crytic/medusa/pull/79/commits/906a4aca25259f4bb03c93a23e14a216e7e5d272)

cmd/completion.go
anishnaik: Resolved in [906a4ac](https://github.com/crytic/medusa/pull/79/commits/906a4aca25259f4bb03c93a23e14a216e7e5d272)

cmd/completion.go
```diff
@@ -0,0 +1,103 @@
+package cmd
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
```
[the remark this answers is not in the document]
anishnaik: Resolved in [906a4ac](https://github.com/crytic/medusa/pull/79/commits/906a4aca25259f4bb03c93a23e14a216e7e5d272)

anishnaik (approved):

Issue #107, opened 2023-03-17
Update error handling


anishnaik:
Currently, we are using the in built error handling of go. The issue with this is that, by default, the errors do not come with stack traces which makes it harder to debug and triage any issues that might arise.

For this reason, we should move to `pkg/errors` that allows you to attach a stack trace at whatever point an error is introduced.

Fixing this issue should not be the final time that we revisit error handling. Down the line we should come up with a set of standardized errors and a set of guidelines on how to approach error handling, where to add additional context, where to attach the error stack, and so on.

Issue #55, opened 2022-12-02
Periodically generate calls to view methods in the CallSequence


Xenomega:
We'll want to periodically call view methods to increase coverage. Additionally, assertion testing supports config-driven options to enable assert testing of view methods, but the `FuzzerWorker` only produces `CallSequence`s with calls targeting state changing methods right now, so it's never hit.

Requirements:
- Record view functions separately, but alongside the existing `stateChangingMethods` (see the `updateStateChangingMethods` method)
- Add a really low probability of sourcing a target method from it (instead of `stateChangingMethods`) at the top of the `generateFuzzedCall` method.
- Ensure the `corpus.UpdateCorpusAndCoverageMaps` method checks if the last call was to a view method. If it was, do not record the call sequence in the corpus at that step, as it's not a coverage-increasing sequence we'd be interested in recording, as it was not state changing.

Jaime-Iglesias:
Will try to pick this up starting tomorrow.

0xalpharush:
>Ensure the corpus.UpdateCorpusAndCoverageMaps method checks if the last call was to a view method. If it was, do not record the call sequence in the corpus at that step, as it's not a coverage-increasing sequence we'd be interested in recording, as it was not state changing.

need clarification on this but https://github.com/crytic/medusa/pull/363 addresses the other points

ggrieco-tob:
I think view functions should be part of the coverage, in particular if they have or can contain assertions.

PR #111, opened 2023-03-24
Coverage report generation

### LINKED ISSUE #40: Coverage reports
Similar to echidna, it will be very valuable to have a "coverage report" that outlines what portions of the code were covered.

These are the key features required:
1. The coverage report should be an HTML file
2. The HTML file should have links for each deployed contract
3. Similar to echidna, we will take the `r`, `*`, `o` methodology. Or come up with a new one if that makes sense
4. Create a `--coverage-report` CLI option that can be called separate from general fuzzing execution

Over time we can add more interesting coverage results or different types of reports. Ideally, we can also have:
1. Show how "often" a line of code was executed
2. TBD

Jaime-Iglesias:
This PR closes #40 .

I have taken the first steps into coverage generation; however, the is still a good chunk of work to do.

Added source code, source lines, parsed runtime bytecode, and parsed runtime source mapping to the contract metadata - this way we don't have to do all of those steps during coverage generation.

https://github.com/trailofbits/medusa/blob/1f9335677647b1ada4d1ecdddba07a8e97f71242/compilation/platforms/solc.go#L161-L189

**Parsing Bytecode:**

Added a method that takes some bytecode (metadata-less is best), identifies each EVM operation in said bytecode, and stores for each operation the start and end offsets it spawns in the bytecode (i.e a PUSH1 instruction that goes from offset 0 to offset 3) , the instruction as a string (tbh this could be removed as I was simply using it for visualization) and a boolean indicating whether the instruction was covered (this bool we can probably remove its left as a previous design decision that I think no longer makes sense).

https://github.com/trailofbits/medusa/blob/4c182697f317b473420dc28fea5b257eebf833ed/compilation/types/compiled_contract.go#L75-L120

**Parsing Metadata:**

Added a method that transforms some given source mapping in the form of a string into a struct containing the expected fields (offset, length, fileID, etc..). It also takes care of handling all the compression of the metadata and transforms those compressed values into the actual ones.

https://github.com/trailofbits/medusa/blob/4c182697f317b473420dc28fea5b257eebf833ed/compilation/types/compiled_contract.go#L122-L197

**Split source code into lines:**

This method simply iterates some given source code and stores the offsets at which each line begins and ends, this will be useful for coverage generation.

https://github.com/trailofbits/medusa/blob/4c182697f317b473420dc28fea5b257eebf833ed/compilation/types/compiled_contract.go#L199-L225

**Coverage Generation:**

I added a `GenerateCoverageReport` report function in the `coverage_generation.go` file. However, this function is incomplete and will need a few changes based on the "what is left to do" part of this PR - basically because we are missing a key step towards coverage generation which are the fileID's generated by the compiler which I have not been able to figure out how to obtain.

There are a lot of comments that explain what the function does and why it does it that way - for example, Medusa stores coverage per address and codehash (e.g map[address] -> map[codeHash] -> coverage) that means that there can be multiple contract definitions that are actually the same contract; therefore we have to aggregate all those contract (we do this through the bytecode hash.

**Template:**

There is a "placeholder code" in `coverage_html.go` containing code related to html templates -  I was using it to test what I was learning from the html templates documentation. It works, it just need the right inputs and an actual html template.

**What is left to do:**

Most of the steps are there (i.e parsed bytecode, parsed source mapping, etc...); However, we are missing some key pieces before we can glue them together into the actual coverage:

1. We need to link each source file to the `id` the compiler generates for them - we should be able to obtain them from the JSON generated by the compiler; however I have not been able to - I have an intuition that we might be dropping that information but I couldn't confirm it through the tests I did.

Here is a sample solc JSON output in which we can see the file id's for reference

image

2. Once we can link `fileID` to source file then what we can do is in the `GenerateCoverageReport` function create a mapping `fileID -> sourceLines` that way when we are iterating over the source mappings we can mark the coverage in the right file based on the information given by element.

3. Creating an html template for the report and filling it with the coverage data.

**Putting it all together**

1. Iterate over all existing contract definitions.
  1.1 Create a map fileID to SourceLines.
  1.2 while we iterate over the definitions figure out a way to link contract definitions that refer to the same contract (I currently do this using the runtime bytecode hash; however, that will not catch all cases).
  
2. Aggregate all the coverage data belonging to the same source file (again this is already done; however, it relies on the runtime bytecode hash which does not catch all cases).

3. Iterate over all contracts which we have aggregated coverage data for
  3.1 Use the aggregated coverage to mark source map elements as covered || not covered (Note that the length of the coverage data is bigger than that of the source mapping, this is because the coverage data is calculated on the bytecode that includes the metadata, therefore use the src mapping as the iterator not the coverage).
  3.2 Once we have marked the elements we can now iterate over the src mapping marking the specific lines in the (fileID -> sourceLines) map as covered.

4. Generate the actual report


compilation/types/source_maps.go
```diff
@@ -0,0 +1,174 @@
...
+	// Separate all the individual source mapping elements
+	elements := strings.Split(sourceMapStr, ";")
+
+	// We use this variable to store "the previous element" because the way
+	// the source mapping works when an element or field is "empty"
+	// the value of the previous element is used.
+	previous := SourceMapElement{
+		Offset:        -1,
+		Length:        -1,
+		FileID:        -1,
```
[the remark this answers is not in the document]
Xenomega: This would largely be a no-op for medusa, with the exception that if a non-empty source map string is supplied without enough "parts" for any element, we might want to throw an error (not allow a valid offset but length of -1). Otherwise, FileID of -1 will be a no-op for how medusa uses it, but for an API consideration, we should also probably throw an error if it's not ever supplied.

I'll resolve this later.

Xenomega:
Remaining TODOs:
- [ ] Reduce coverage lines to the smallest (leaves) of source mapped objects, so we don't include source mappings that include an entire function (happens on entry).
- [ ] Store reverted paths in coverage maps
- [ ] Make the corpus save sequences which increased reverting coverage (in a separate folder)
- [ ] Move call sequences that triggered test failures into a seperate folder in corpus (#114)
- [ ] Ensure constructor coverage is appropriately mapped/collected.
- [ ] Call view functions periodically (#55)

Less important, to do later:
- [ ] Make the report HTML template nicer.

compilation/platforms/solc.go
```diff
@@ -104,6 +104,18 @@ func (s *SolcCompilationConfig) Compile() ([]types.Compilation, string, error) {
 
 	// Create a compilation unit out of this.
 	compilation := types.NewCompilation()
+	if sourceList, ok := results["sourceList"]; ok {
+		if sourceListCasted, ok := sourceList.([]any); ok {
```
anishnaik, line 108: Why don't we assert `sourceList.([]string)`? In that way the errors make more sense and we don't have to "worry" about `sourceListCasted[i].(string)` causing some unlikely panic.

Xenomega: You can't do this because of how JSON marshaling and Go types handle these sorts of things. Although the elements of this `[]any` slice are of type `string`, the overarching array type is defined in Go as `[]any` by the `json` package. You can't cast a `[]any` to `[]string` in Go. We have a helper util for this I believe, but you still need to cast to `[]any` first.

We could add a check on the string casting though. Realistically it would be better to rewrite all of this to unmarshal structs in the future though, like the crytic platform code.

anishnaik: Smh golang...thank you for the explanation.

compilation/types/source_maps.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	// Return the resulting map
+	return sourceMap, nil
+}
+
+// GetInstructionIndexToOffsetLookup obtains a slice where each index of the slice corresponds to an instruction index,
+// and the element of the slice represents the instruction offset.
+// Returns the slice lookup, or an error if one occurs.
+func (s SourceMap) GetInstructionIndexToOffsetLookup(bytecode []byte) ([]int, error) {
```
anishnaik, line 153: Doesn't really matter what the choice here is but does this have to be a function of `SourceMap`? It seems like the only thing that is needed from the `SourceMap` is its length. Seems like this can just be a helper function since no aspect of the `SourceMap` is used / changed

fuzzing/fuzzer_test_methods_test.go
```diff
@@ -80,14 +80,17 @@ func assertFailedTestsExpected(f *fuzzerTestContext, expectFailure bool) {
 // assertCorpusCallSequencesCollected will check to see whether we captured coverage-increasing call sequences in the
 // corpus. It asserts that the actual result matches the provided expected result.
 func assertCorpusCallSequencesCollected(f *fuzzerTestContext, expectCallSequences bool) {
+	// Obtain our count of mutable (often representing just non-reverted coverage increasing) sequences.
+	callSequenceCount := f.fuzzer.corpus.CallSequenceEntryCount(true, false, false)
```
anishnaik, line 84: Feel like the better choice of function here is `f.fuzzer.corpus.ActiveMutableSequenceCount`

anishnaik: Or maybe those two values can differ?

compilation/platforms/solc.go
```diff
@@ -104,6 +104,18 @@ func (s *SolcCompilationConfig) Compile() ([]types.Compilation, string, error) {
...
+			compilation.SourceList = make([]string, len(sourceListCasted))
+			for i := 0; i < len(sourceListCasted); i++ {
+				compilation.SourceList[i] = sourceListCasted[i].(string)
+			}
+		} else {
+			return nil, "", fmt.Errorf("could not parse compiled source artifact because 'sourcesList' was not a []string type")
```
Jaime-Iglesias, line 114: Maybe we could make this error message a bit more informative? hahaha

Perhaps we should say something like "incorrect JSON type for `sourcesList`, was expecting a []string"

compilation/platforms/truffle.go
```diff
@@ -65,6 +65,8 @@ func (s *TruffleCompilationConfig) Compile() ([]types.Compilation, string, error
 	// Create a compilation unit out of this.
 	compilation := types.NewCompilation()
 
+	// TODO: Parse a source list!
```
Jaime-Iglesias, line 68: Is this meant to be left as a TODO ?

fuzzing/fuzzer_test.go
```diff
@@ -590,8 +590,10 @@ func TestCorpusReplayability(t *testing.T) {
 			assertCorpusCallSequencesCollected(f, true)
 			newCoverage := f.fuzzer.corpus.CoverageMaps()
 
-			// Check to see if original and new coverage are the same
-			assert.True(t, originalCoverage.Equals(newCoverage))
+			// Check to see if original and new coverage are the same.
+			// TODO: Fix this, as coverage now tracks reverted too, and if this test solved too quickly on first run,
```
Jaime-Iglesias, line 594: was this meant to leave as a TODO ?

fuzzing/coverage/coverage_maps.go
```diff
@@ -40,55 +43,134 @@ func NewCoverageMaps() *CoverageMaps {
...
+
+// getContractCoverageMapHash obtain the hash used to look up a given contract's ContractCoverageMap.
+// If this is init bytecode, metadata and abi arguments will attempt to be stripped, then a hash is computed.
+// If this is runtime bytecode, the metadata ipfs/swarm hash will be used if available, otherwise the bytecode
+// is hashed.
+// Returns the resulting lookup hash.
+func getContractCoverageMapHash(bytecode []byte, init bool) common.Hash {
+	// If available, the metadata code hash should be unique and reliable to use above all (for runtime bytecode).
+	if !init {
+		metadata := compilationTypes.ExtractContractMetadata(bytecode)
```
anishnaik, line 85: Should we panic / throw an error if `metadata` is `nil`?

anishnaik: nvm ignore - this case is handled.

fuzzing/coverage/coverage_tracer.go
```diff
@@ -146,28 +148,16 @@ func (t *CoverageTracer) CaptureState(pc uint64, op vm.OpCode, gas, cost uint64,
 
 	// If there is code we're executing, collect coverage.
 	if len(scope.Contract.Code) > 0 {
-		// We record coverage maps under a code hash to merge coverage across different deployments of a contract.
```
anishnaik, line 149: So the above line checks if the thing we are calling has code. Wouldn't the precompiles bypass this check as well? Does it matter?

Xenomega: I'm not even sure `CaptureState` is called on pre-compiles, but if it's empty in any case, this is fine, we can't collect anything from a pre-compile.

anishnaik (approved):
lgtm. tested across various complex codebases and it seems to be working properly.

compilation/platforms/solc.go
Xenomega: Agreed, though this applies in a few areas and all the error handling in these files does this. We nuked `truffle` support in this PR, and we might collapse `solc` support and just have one config that's crytic-compile based moving forward, until we re-evaluate in the future, so maybe it's best I'll leave this one for another time.

Anish's error handling PR might be a good one to touch this up in. I'll reference it so we don't lose track of it there (#107)

compilation/platforms/truffle.go
Xenomega: This was removed in https://github.com/crytic/medusa/pull/111/commits/a83955a3df63d4b15938c029150a54e05d72ec83

fuzzing/fuzzer_test.go
Xenomega: This was removed in https://github.com/crytic/medusa/pull/111/commits/9e77c357bb5c06abd8992863c3495922b08fc674

Xenomega:
This PR is ready to go. I made some HTML/CSS tweaks that shouldn't have broken anything. I tested it on Ubuntu/Windows, and will test it on mac OS tomorrow before merging.

PR #122, opened 2023-04-05
Standarize headless mode status information

ggrieco-tob:
This PR should help to addresses at least some of the tasks in #121


anishnaik:
headless mode will not be implemented in the short term. we can reopen this PR if necessary.

PR #135, opened 2023-04-19
Fix internal library event emissions in ExecutionTrace

### LINKED ISSUE #131: Bug: medusa fails to decode events emitted in internal library functions
```
library Debugger {
    event Debug(string debugString);

    function log(string memory debugString) internal {
        emit Debug(debugString);
    }
}

contract TestContract {
    function testMedusaEvent() external {
        Debugger.log("test");
        assert(false);
    }
}
```

In the following code, the event is emitted by the `Debugger.log` function but when we try to match event logs to an ABI, we are matching it with `TestContract`'s ABI instead of `Debugger`'s ABI so we are able to unresolve it and get output like this:

```
[FAILED] Assertion Test: Medusa.testMedusaEvent()
Test for method "Medusa.testMedusaEvent()" failed after the following call sequence resulted in an assertion:
1) Medusa.testMedusaEvent() (block=2, time=372470, gas=12500000, gasprice=1, value=0, sender=0x0000000000000000000000000000000000010000)
 -> [Execution Trace]
 -> [call] Medusa.testMedusaEvent() (addr=0xA647ff3c36cFab592509E13860ab8c4F28781a66, value=0, sender=0x0000000000000000000000000000000000010000)
         -> [event] <unresolved(topics=[7cdb51e9dbbc205231228146c3246e7f914aa6d4a33170e43ecc8e3593481d1a], data=000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000047465737400000000000000000000000000000000000000000000000000000000)>
         -> [assertion failed]
```

Xenomega:
This PR closes #131 with a temporary fix.

The issue was that events emitted through internal libraries were not being resolved. This provides a temp fix where if the immediate contract ABI does not contain the event definition, all contract definitions will be iterated over to find a matching event.

A future fix should include discovering associations between contracts and libraries and only resolving the event from relevant associated libraries.

However, for now, I'm confident this should work without any practical issues popping up.

Changes:
- `ExecutionTrace` now stores a reference to all contract definitions
- `ExecutionTracer` initializes `ExecutionTrace` with all contract definitions
- `ExecutionTrace` uses all contract definitions as a backup to try and resolve _any_ event which matches the current one, if the immediate contract ABI did not contain the correct event definition.
- Updated event emission test to include use of an internal library.

### DIFF
--- a/fuzzing/executiontracer/execution_trace.go
+++ b/fuzzing/executiontracer/execution_trace.go
@@ -201,8 +201,9 @@ func (t *ExecutionTrace) generateEventEmittedString(callFrame *CallFrame, eventL
 	// Try to unpack our event data
 	event, eventInputValues := abiutils.UnpackEventAndValues(callFrame.CodeContractAbi, eventLog)
 	if event == nil {
-		// TODO: If we couldn't resolve the event, it comes from a library. Temporarily, we just try to resolve the
-		//  event from all contracts, until we can associate which contracts actually represent relevant libraries.
+		// If we couldn't resolve the event from our immediate contract ABI, it may come from a library.
+		// TODO: Temporarily, we fix this by trying to resolve the event from any contracts definition. A future
+		//  fix should include only checking relevant libraries associated with the contract.
 		for _, contract := range t.contractDefinitions {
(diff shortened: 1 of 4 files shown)


anishnaik (approved):
lgtm

PR #143, opened 2023-05-02
Add Nix support

bohendo:
Once merged to the default branch, this PR should allow anyone with nix installed to run `nix run github:crytic/medusa` to use the medusa tool without installing anything globally. Kind of like a single-serving virtualenv. The first time this is run, it'll take some time as medusa's dependencies (eg `slither`) are downloaded & the project is built, but subsequent runs will start almost instantly without requiring any further downloads. [More info re `nix run`](https://determinate.systems/posts/nix-run).

Allows global installation of medusa with: `nix build && nix-env -i ./result`. This `medusa` installation is hooked up to a new copy of required dependencies, so it'll take care of the slither dependency if not already present but w/out any risk of conflict w existing global `slither` installation.

Also provides a developer environment via `nix develop` which opens a shell that includes basic golang tools & other dependencies needed to build & run tests.

This is entirely opt-in, it uses the rest of the repo's config (eg `go.mod`) rather than attempting to replace it. If you don't want to fiddle with `nix`, then you can ignore these new files & do the same thing you've always done.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,7 +2,7 @@
   description = "Medusa smart-contract fuzzer";
 
   inputs = {
-    nixpkgs.url = "nixpkgs/nixos-22.11";
+    nixpkgs.url = "nixpkgs/nixos-24.11";
     flake-utils.url = "github:numtide/flake-utils";
   };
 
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -0,0 +1,128 @@
...
+            src = builtins.fetchGit {
+              url = "git+ssh://git@github.com/crytic/solc-select";
+              rev = "97f160611c39d46e27d6f44a5a61344e6218d584";
+            };
+            propagatedBuildInputs = with pkgs.python39Packages; [
+              packaging
+              setuptools
+              pycryptodome
+            ];
+          });
```
bohendo, line 36: In-line addition of nix support to `solc-select` & other packages that don't feature a `flake.nix` file yet. If/when they do, these pieces of config can move into their respective repos & can be removed from this project.
In the meantime, they allow `medusa` to depend on a specific commit of `crytic-compile`, etc

bsamuels453 (approved):
lgtm, over the long term we should add nix to our to-be-created release pipeline to ensure the vendorhash is correct

PR #147, opened 2023-05-09
Value shrinking

ggrieco-tob:
This PR adds code to perform a simple value shrinking. It will try to shrik the complexity of the parameters of important transaction, using some simple shirking mutations (200 times per transaction). The code will need additional documentation, tests and a few rounds of review before, but I think it is better to start early.


fuzzing/valuegeneration/mutator_shrinking.go
```diff
@@ -0,0 +1,207 @@
...
+	return b
+}
+
+// integerShrinkingMethods define methods which take a big integer and a set of inputs and
+// transform the integer with a random input and operation. This is used in a loop to create
+// mutated integer values.
+var integerShrinkingMethods = []func(*ShrinkingValueMutator, *big.Int, ...*big.Int) *big.Int{
+	func(g *ShrinkingValueMutator, x *big.Int, inputs ...*big.Int) *big.Int {
+		// Subtract a random input
+		var r *big.Int
```
anishnaik, line 111: Why don't we do `r := big.NewInt(0)` compare `x` with `r` and then get rid of the `else` and just return `r` if the `if` and the `else if` don't hit?

fuzzing/valuegeneration/mutator_shrinking.go
```diff
@@ -0,0 +1,207 @@
...
+		if x.Cmp(big.NewInt(0)) > 0 {
+			r = big.NewInt(0).Sub(x, inputs[g.randomProvider.Intn(len(inputs))])
+		} else if x.Cmp(big.NewInt(0)) < 0 {
+			r = big.NewInt(0).Add(x, inputs[g.randomProvider.Intn(len(inputs))])
+		} else {
+			r = big.NewInt(0)
+		}
```
anishnaik, line 118: Being nitpicky - Can we add a comment for each conditional branch?

fuzzing/valuegeneration/mutator_shrinking.go
```diff
@@ -0,0 +1,207 @@
...
+		return r
+
+	},
+	func(g *ShrinkingValueMutator, x *big.Int, inputs ...*big.Int) *big.Int {
+		// Divide by two
+		return big.NewInt(0).Div(x, big.NewInt(2))
+	},
+	func(g *ShrinkingValueMutator, x *big.Int, inputs ...*big.Int) *big.Int {
+		return big.NewInt(0)
```
anishnaik, line 127: Add comment here to say "Return zero"

fuzzing/valuegeneration/mutator_shrinking.go
```diff
@@ -0,0 +1,207 @@
...
+		return input
+	}
+	return b
+}
+
+// integerShrinkingMethods define methods which take a big integer and a set of inputs and
+// transform the integer with a random input and operation. This is used in a loop to create
+// mutated integer values.
+var integerShrinkingMethods = []func(*ShrinkingValueMutator, *big.Int, ...*big.Int) *big.Int{
+	func(g *ShrinkingValueMutator, x *big.Int, inputs ...*big.Int) *big.Int {
```
anishnaik, line 109: Can we add a comment for each `integerShrinkingMethod` explaining what it does? Similar to the format for `bytesShrinkingMethods`

fuzzing/valuegeneration/mutator_shrinking.go
```diff
@@ -0,0 +1,207 @@
...
+			return b
+		}
+
+		i := g.randomProvider.Intn(len(b))
+		return append(b[:i], b[i+1:]...)
+	},
+}
+
+// MutateBytes takes a dynamic-sized byte array input and returns a mutated value based off the input.
+func (g *ShrinkingValueMutator) MutateBytes(b []byte) []byte {
```
anishnaik, line 95: Why doesn't byte shrinking or string shrinking go through rounds like integer shrinking?

fuzzing/valuegeneration/mutator_shrinking.go
```diff
@@ -0,0 +1,207 @@
...
+
+		// Correct value boundaries (underflow/overflow)
+		input = utils.ConstrainIntegerToBounds(input, min, max)
+		return input
+	}
+	return i
+}
+
+// stringShrinkingMethods define methods which take an initial string and a set of inputs to transform the input. The
+// transformed input is returned. This is used in a loop to mutate strings.
```
anishnaik, line 171: Is the function used in a loop? This pertains to the last sentence of the comment

fuzzing/valuegeneration/mutator_shrinking.go
```diff
@@ -0,0 +1,207 @@
...
+		// in our set, and we don't want duplicates.
+		var inputs []*big.Int
+		inputs = append(inputs, g.valueSet.Integers()...)
+		if signed {
+			inputs = append(inputs, min, max)
+		} else {
+			inputs = append(inputs, max)
+		}
+
+		// Determine which value we'll use as an initial input, and how many mutations we will perform.
```
anishnaik, line 150: Remove "how many mutations we will perform" since we are not doing rounds of shrinking

fuzzing/fuzzer_worker.go
```diff
@@ -301,105 +319,149 @@ func (fw *FuzzerWorker) testCallSequence() (calls.CallSequence, []ShrinkCallSequ
...
+			optimizedSequence = possibleShrunkSequence
 		} else {
 			// We didn't remove an item at this index, so we'll iterate to the next one.
 			i++
 		}
 	}
 
+	// Next try to shrink our values of every transaction a given number of rounds.
+	for i := 0; i < len(optimizedSequence); i++ {
+		for optimizationRound := 0; optimizationRound < 200; optimizationRound++ {
```
anishnaik, line 423: I believe echidna does 5000 shrink rounds by default - is there a reason we are shooting for 200?

anishnaik: Do we want to add a config option for it?

fuzzing/valuegeneration/mutator_shrinking.go
Xenomega: Addressed in https://github.com/crytic/medusa/pull/147/commits/dd31ac5546a8dd1f41e54d03157e1165213fcbd8

fuzzing/valuegeneration/mutator_shrinking.go
Xenomega: Addressed in https://github.com/crytic/medusa/pull/147/commits/dd31ac5546a8dd1f41e54d03157e1165213fcbd8

fuzzing/valuegeneration/mutator_shrinking.go
Xenomega: https://github.com/crytic/medusa/pull/147/commits/dd31ac5546a8dd1f41e54d03157e1165213fcbd8

fuzzing/fuzzer_worker.go
Xenomega: What do you think? So basically, we try to optimize values for each call 200 times. So the work will be 200 * call sequence length.

I was thinking we should update this because the value shrinker uses probability to determine when to shrink, so we should always force it to shrink at least one value in the future (updating `abi_values.go`). So for now I was thinking this could be fixed because it's low and probably won't be a problem, but maybe it still would be good to add an option? I can quickly add it if you think so.

fuzzing/valuegeneration/mutator_shrinking.go
Xenomega: Updated in https://github.com/crytic/medusa/pull/147/commits/c89252203c403f26938e1375af059451a3469d77

fuzzing/fuzzer_worker.go
anishnaik: I think if we are going to do another pass around the shrinking logic anyway, we can keep as is. i don't think it matters much either way. We can see the efficacy of 200 in practice and evaluate whether we want to increase, decrease, or allow the user the flexibility to choose.

anishnaik (approved):
bing bong

PR #158, opened 2023-06-13
ci: avoid using C: on Windows CI

elopez:
The I/O on the C: drive is particularly slow on GitHub Actions. Configure
venvs and cache locations to point to the D: (workspace) drive to improve
the execution time of the CI workflow.

This also updates the `setup-go` action to v4, which includes caching for dependencies, resulting in faster runs for Ubuntu and macOS. It also adds timeouts to jobs, to avoid wasting minutes if jobs get stuck.

The execution times can vary wildly from run to run, but I believe this is an overall improvement in times.

Before:
Screenshot 2023-06-13 at 14 32 40
Screenshot 2023-06-13 at 14 38 18

After: 
Screenshot 2023-06-13 at 14 35 02
Screenshot 2023-06-13 at 14 37 54

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,8 @@ jobs:
         environment: [ubuntu-latest, macos-latest, windows-latest]
 
     runs-on: ${{ matrix.environment }}
+    timeout-minutes: 10
+
     steps:
       - uses: actions/checkout@v3
 
@@ -69,6 +71,7 @@ jobs:
 
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -105,17 +153,15 @@ jobs:
...
 
       - name: Install Node dependencies
-        run: npm install -g hardhat
+        run: npm install hardhat
 
       - name: Install Python dependencies
         run: |
-          pip3 install solc-select
-          pip3 install slither-analyzer
+          pip3 install --no-cache-dir solc-select slither-analyzer
```
anishnaik, line 160: It's still uncertain whether we will integrate slither with medusa so can we change `slither-analyzer` to `crytic-compile`?

anishnaik (approved):

Issue #142, opened 2023-05-02
Extend assertion-testing with different panic codes


Xenomega:
Currently, the assertion testing provider only triggers a failed assertion if the panic code is `AssertFailed` (indicating an actual Solidity `assert(...)` statement panicked).

We should extend the `AssertionTestingConfig` to also have boolean options for other types of panics, such as overflows. By default, these extended options should be disabled so only `AssertFailed` is enabled by default. We should also add a config option for `AssertFailed` itself, e.g. if someone wants to only test overflows.

References:
- [Assertion-testing logic](https://github.com/crytic/medusa/blob/b7ea61bd76cbdb0520a5c4d34c7f1a8dc139f54e/fuzzing/test_case_assertion_provider.go#L76) (to be changed)
- [Solidity panic codes](https://github.com/crytic/medusa/blob/b7ea61bd76cbdb0520a5c4d34c7f1a8dc139f54e/compilation/abiutils/solidity_errors.go) (to be leveraged)

TODOs:
- Update `config.go` and `config_defaults.go` with the new config options.
- Update `assertion_test_case_provider.go` in the spot linked above, to check the `Fuzzer`'s config for these options and perform the appropriate checks there.
- Add unit tests to `fuzzer_test.go` to test the different panic conditions enabled/disabled working accordingly.

anishnaik:
This issue has been resolved. Closing.

PR #161, opened 2023-06-20
Extend assertion tests using different panic codes

damilolaedwards:
This PR aims to extend assertion-testing with different panic codes using options passed in via the config file. [[full details]](https://github.com/crytic/medusa/issues/142)


fuzzing/config/config.go
```diff
@@ -113,6 +113,9 @@ type AssertionTestingConfig struct {
 
 	// TestViewMethods dictates whether constant/pure/view methods should be tested.
 	TestViewMethods bool `json:"testViewMethods"`
+
+	// AssertOptions describes options for testing panics using the panic codes.
+	AssertOptions []byte `json:"assertOptions"`
```
anishnaik, line 118: So after some thought I don't think we want a `[]byte` for this. The reason being it sort of diminishes the user experience since the user will have to go research what the different panic codes are and what their hex representations are. The better way would be to create an object that can individually enable each panic code.

e.g.

```
type AssertionTestingConfig struct {
...
AssertionModes AssertionModesConfig `json:"assertionModes"`
...
}

type AssertionModesConfig struct {
	FailOnArithmeticUnderflow bool `json:"failOnArithmeticUnderflow"`
}
```

This way the user can just set things to true and false. In config_defaults, we can have `FailOnAssertion` as the only one as `true` and the rest can be false.

Xenomega: +1 to what Anish said. As a note, I'd just re-reference the issue associated to this, as it provided a list to medusa panic code definitions to leverage and mentioned designing this with bools instead. It might provide you more useful context here. We'd want one config option (bool) per panic code in medusa in the end.

fuzzing/test_case_assertion_provider.go
```diff
@@ -73,7 +73,7 @@ func (t *AssertionTestCaseProvider) checkAssertionFailures(callSequence calls.Ca
 	// have a panic code.
 	lastExecutionResult := lastCall.ChainReference.MessageResults().ExecutionResult
 	panicCode := abiutils.GetSolidityPanicCode(lastExecutionResult.Err, lastExecutionResult.ReturnData, true)
-	encounteredAssertionFailure := panicCode != nil && panicCode.Uint64() == abiutils.PanicCodeAssertFailed
+	encounteredAssertionFailure := panicCode != nil && utils.IsPanicCodeIncluded(byte(panicCode.Uint64()), t.fuzzer.config.Fuzzing.Testing.AssertionTesting.AssertOptions) && utils.HasEncounteredAssertionFailure(panicCode)
```
anishnaik, line 76: So this is the right idea but I think we can remove the dependency with `assertion_utils` by storing which panic codes are enabled/disabled directly in the provider. I would look at the `attachAssertionTestCaseProvider` function and see how you can instantiate some structure that would hold which panic codes are enabled. Note that the `attachAssertionTestCaseProvider` takes in a `Fuzzer` object so you have the underlying `Fuzzing.Config` at your disposal. My initial thoughts are to create a mapping (maybe `map[byte]bool` ? but not 100% sure) that you can simply query to see if some `panicCode` is enabled or not

anishnaik: So the question is how do you take the `AssertionModeConfig` object, store it in a meaningful way in the `AssertionTestCaseProvider` object, instantiate that object correctly in `attachAssertionTestCaseProvider` and use it accoredingly in `checkAssertionFailures`

Xenomega: If you're switching to individual config options you'll need to reference, you can create this mapping in the test case provider on fuzzer start.

Otherwise you can just make it a switch/if-else statement. That may actually be computationally faster than a dictionary lookup given the small number of items and that it's a single byte compare.

fuzzing/testdata/contracts/assertions/assert_allocate_too_much_memory.sol
```diff
@@ -0,0 +1,6 @@
+contract TestContract {
```
anishnaik, line 1: For each solidity test file, can you add a comment at the top of it to highlight what the contract is doing? So for this one here is an example:

```
This contract ensures that the fuzzer will fail if too much memory is allocated
```

anishnaik:
Can we also make sure to merge with master after we resolve all the above concerns?

fuzzing/config/config.go
```diff
@@ -106,13 +106,28 @@ type TestingConfig struct {
...
+	FailOnAssertFailed                  bool `json:"fail_on_assert_failed"`
+	FailOnArithmeticOverflow            bool `json:"fail_arithmetic_overflow"`
+	FailOnDivideByZero                  bool `json:"fail_divide_by_zero"`
+	FailOnEnumTypeConversionOutOfBounds bool `json:"fail_on_enum_type_conversion_out_of_bounds"`
+	FailOnIncorrectStorageAccess        bool `json:"fail_on_incorrect_storage_access"`
+	FailOnPopEmptyArray                 bool `json:"fail_on_pop_empty_array"`
+	FailOnOutOfBoundsArrayAccess        bool `json:"fail_on_out_of_bounds_array_access"`
+	FailOnAllocateTooMuchMemory         bool `json:"fail_on_allocate_too_much_memory"`
+	FailOnCallUninitializedVariable     bool `json:"fail_on_call_uninitialized_variable"`
+}
```
anishnaik, line 119: Can you add a comment for the top-level `AssertionModesConfig` as well as all its parameters? You can use the various structs in the config files as an example.

fuzzing/fuzzer_test.go
```diff
@@ -81,6 +81,47 @@ func TestAssertionsBasicSolving(t *testing.T) {
 	}
 }
 
+// TestAssertionsBasicSolving runs tests to ensure that assertion testing behaves as expected.
+func TestAllPanicAssertions(t *testing.T) {
```
anishnaik, line 85: Update the test to be called something like `TestAssertionModes` and update the comment above to match it

fuzzing/testdata/contracts/assertions/assert_allocate_too_much_memory.sol
```diff
@@ -0,0 +1,9 @@
+//This contract attempts to allocate an excessive amount of memory by creating an array with a length of 2^64 causing a panic.
+//PanicCodeAllocateTooMuchMemory = 0x41
```
anishnaik, line 2: Can we make sure to add spaces after the "//" and before the comment. Sorry for being nitpicky but just want to make sure we follow the established pattern

anishnaik: Same applies to all the test solidity files

fuzzing/test_case_assertion_provider.go
```diff
@@ -73,7 +73,44 @@ func (t *AssertionTestCaseProvider) checkAssertionFailures(callSequence calls.Ca
...
+				// Get the panic code associated with the config option
+				code, found := panicCodes[fieldName]
+
+				// Check if the panic code matches the given panic code
+				if found && code == panicCode.Int64() {
+					encounteredAssertionFailure = true
+				}
+			}
+		}
+	}
```
anishnaik, line 113: I think this overcomplicates the situation. Ideally, we maintain what panic modes are enabled within `AssertionTestCaseProvider`

anishnaik: Even simpler would be to have a switch statement here like Xenomega mentioned. Will write the pseudocode

anishnaik: Here is some pseudo-code that you can add to `attachAssertionTestCaseProvider`

```
if config.FailOnAssertion {
    t.panicCodes[abiutils.PanicCodeAssertFailed] = true
}
if config.FailOnUnderflow {
   t.panicCodes[abiutils.PanicCodeUnderflow] = true
}
```

Now in `checkAssertionFailure` you just have to do a key-value lookup in `AssertionTestCaseProvider.panicCodes`.  So `if t.panicCodes[panicCode] == true` then we can set `encounteredAssertionFailure` to true.

anishnaik:
@damilolaedwards we will also need to update this snippet of code as well
https://github.com/crytic/medusa/blob/5a3c83965f2ef953e3cf4f6b99679674ec13abc5/fuzzing/executiontracer/execution_trace.go#L166

anishnaik:
Ready for your review @Xenomega

anishnaik:
This PR closes #161

fuzzing/test_case_assertion_provider.go
```diff
@@ -234,3 +235,79 @@ func (t *AssertionTestCaseProvider) callSequencePostCallTest(worker *FuzzerWorke
...
+// panic code that was hit should be treated as a failing case - which will be determined by whether that panic
+// code was enabled in the config. Note that the panic codes are defined in the abiutils package and that this function
+// panic if it is provided a panic code that is not defined in the abiutils package.
+// TODO: Also note that this is a terrible design and if there is a better way to maintain assertion and panic logic, it
+//
+//	should be implemented in a future PR
+func encounteredAssertionFailure(panicCode uint64, conf config.AssertionModesConfig) bool {
+	// Switch on panic code
+	switch panicCode {
+	case abiutils.PanicCodeCompilerInserted:
```
Xenomega, line 249: Here I'd simplify this to just `return conf.<option>`. The if statements shouldn't be necessary

fuzzing/test_case_assertion_provider.go
```diff
@@ -234,3 +235,79 @@ func (t *AssertionTestCaseProvider) callSequencePostCallTest(worker *FuzzerWorke
 
 	return shrinkRequests, nil
 }
+
+// encounteredAssertionFailure takes in a panic code and a config.AssertionModesConfig and will determine whether the
+// panic code that was hit should be treated as a failing case - which will be determined by whether that panic
+// code was enabled in the config. Note that the panic codes are defined in the abiutils package and that this function
+// panic if it is provided a panic code that is not defined in the abiutils package.
+// TODO: Also note that this is a terrible design and if there is a better way to maintain assertion and panic logic, it
```
Xenomega, line 243: Comment spacing fix here

fuzzing/testdata/contracts/assertions/assert_allocate_too_much_memory.sol
```diff
@@ -0,0 +1,9 @@
+// This contract attempts to allocate an excessive amount of memory by creating an array with a length of 2^64 causing a panic.
+// PanicCodeAllocateTooMuchMemory = 0x41
+
+contract TestContract {
+    function allocateTooMuchMemory() public {
+        uint256[] memory myArray = new uint256[](2**64); // Allocate too much memory
+        myArray[2**64 - 1] = 42;
+    }
+}
```
Xenomega, line 9: Small nitpick, but I'd add new lines to the end of all these solidity tests so git doesn't warn about it.

compilation/abiutils/solidity_errors.go
```diff
@@ -105,3 +105,35 @@ func GetSolidityCustomRevertError(contractAbi *abi.ABI, returnError error, retur
 	}
 	return nil, nil
 }
+
+// GetPanicReason will take in a panic code as an uint64 and will return the string reason behind that panic code. For
+// example, if panic code is PanicCodeAssertFailed, then "assertion failure" is returned. This is used by the execution
```
Xenomega, line 110: Small nitpick, but with this comment, I wouldn't mention what it's used for here, since this is supposed to be agnostic from the fuzzing package, and it can be used for anything.

compilation/abiutils/solidity_errors.go
```diff
@@ -105,3 +105,35 @@ func GetSolidityCustomRevertError(contractAbi *abi.ABI, returnError error, retur
...
+	case PanicCodePopEmptyArray:
+		return "pop on empty array"
+	case PanicCodeOutOfBoundsArrayAccess:
+		return "out of bounds array access"
+	case PanicCodeAllocateTooMuchMemory:
+		return "overallocation of memory"
+	case PanicCodeCallUninitializedVariable:
+		return "call on uninitialized variable"
+	default:
+		// TODO: Should we throw a panic here or just return this string?
```
Xenomega, line 136: I'd say here we can return a string that is more like "unknown panic code(<code>)", so its still clear what it is.

fuzzing/test_case_assertion_provider.go
```diff
@@ -234,3 +235,79 @@ func (t *AssertionTestCaseProvider) callSequencePostCallTest(worker *FuzzerWorke
...
+	case abiutils.PanicCodeCallUninitializedVariable:
+		if conf.FailOnCallUninitializedVariable {
+			return true
+		} else {
+			return false
+		}
+	}
+
+	// Throw a panic if we encounter a panic code not defined in the switch statement above
+	panic("encountered unknown panic code")
```
Xenomega, line 312: Oh last thing, we probably don't need to panic here. If solidity adds a new panic code, we should be mindful of it to keep an eye out (hopefully the "unknown error code" message will help), but I'd say we ignore it and keep fuzzing.

I'd only panic if there was some bad code on our end where we defined some panic code type but never created a string for it, but that isn't a problem to worry about in this implementation really.

anishnaik: Resolved in [cd27dcb](https://github.com/crytic/medusa/pull/161/commits/cd27dcb384e4f0ae2dc52f67534a6b329b1c1ec1)

compilation/abiutils/solidity_errors.go
anishnaik: Resolved in [cd27dcb](https://github.com/crytic/medusa/pull/161/commits/cd27dcb384e4f0ae2dc52f67534a6b329b1c1ec1)

compilation/abiutils/solidity_errors.go
anishnaik: Resolved in [cd27dcb](https://github.com/crytic/medusa/pull/161/commits/cd27dcb384e4f0ae2dc52f67534a6b329b1c1ec1)

fuzzing/test_case_assertion_provider.go
anishnaik: Resolved in [cd27dcb](https://github.com/crytic/medusa/pull/161/commits/cd27dcb384e4f0ae2dc52f67534a6b329b1c1ec1)

fuzzing/test_case_assertion_provider.go
anishnaik: wow, that was dumb lol

resolved in [cd27dcb](https://github.com/crytic/medusa/pull/161/commits/cd27dcb384e4f0ae2dc52f67534a6b329b1c1ec1)

Xenomega (approved):

PR #162, opened 2023-06-20
fix isPropertyTest function

anishnaik:
The `isPropertyTest` function identifies whether a given ABI method is a property test or not. However, this function failed to check whether the function was read-only. Thus, the following function would have been considered valid

```
function my_func() returns(bool) public {
  // I am changing state
 return true;
}
```

The fix is to simply add a `method.IsConstant` check.

Note that this PR also fixes up a few comments.


anishnaik:
It seems like echidna allows for non-view property tests. So to maintain parity this PR will be closed.

PR #165, opened 2023-06-22
Add a parameter in optimization mode to maximize gas usage

ggrieco-tob:


dguido:
This PR is 2.5 years old with significant codebase changes since then. Closing as stale, but the gas optimization feature idea is valuable. @gustavo-grieco - if you're still interested, a fresh PR against the current codebase would be welcome.

Issue #132, opened 2023-04-12
Add `CallAllContracts` option to only call contracts specified in `DeploymentOrder`


ggrieco-tob:
This example should NOT be solved if `allContracts` is false, since medusa should be allowed to call `C` (and not `A`). 

```solidity
interface iA {
  function produceFail() external;
}

contract A {
  function fail(address a) public {
    iA(a).produceFail();
  }
}


contract C {
  bool state = true;
  A a = new A();

  function pass() public {  assert(state); }
  function produceFail() public {
     require(msg.sender == address(a)); 
     state = false;
  } 
} 
```

However, using the command line:

```
$ medusa fuzz --target simple.sol --deployment-order C --assertion-mode
[FAILED] Assertion Test: C.pass()
Test for method "C.pass()" failed after the following call sequence resulted in an assertion:
1) A.fail(0xa647ff3c36cfab592509e13860ab8c4f28781a66) (block=3065, time=360624, gas=12500000, gasprice=1, value=0, sender=0x0000000000000000000000000000000000030000)
2) C.pass() (block=40515, time=721248, gas=12500000, gasprice=1, value=0, sender=0x0000000000000000000000000000000000030000)
 -> [Execution Trace]
 -> [call] C.pass() (addr=0xA647ff3c36cFab592509E13860ab8c4F28781a66, value=0, sender=0x0000000000000000000000000000000000030000)
	 -> [assertion failed]
```

Even using a medusa.json file, with `allContracts` set to false will not work.

anishnaik:
I actually don't think `TestAllContracts` is implemented yet - I can take a deeper look tomorrow.

Xenomega:
` TestAllContracts` is implemented correctly here I believe. We only run tests in `C`, but the way we chose to implement this, `medusa` will call _any_ contract it knows of existing to try and violate the property. This way medusa can find all the ways a test can be violated even indirectly.

I think maybe this issue should be translated into one like:

"Implement a `CallAllContracts` feature similar to `TestAllContracts` that allows for disabling of calls to non-deployment order contracts."

ggrieco-tob:
> "Implement a CallAllContracts feature similar to TestAllContracts that allows for disabling of calls to non-deployment order contracts." so you can test using a different flow if you wanted to.

This is correct.

ggrieco-tob:
I will start working on this to assign a new config option that replicates echidna behavior. After that, we can decide what to do exactly.

ggrieco-tob:
[...]

Issue #51, opened 2022-12-01
High level testing API


Xenomega:
We currently have a good way to extend testing methodology using a lower level API. We'll want to add methods like `RegisterTestCase` and `ReportTestCaseFinished` into a higher level testing suite object that also wraps `CallSequenceTestFunc` and makes writing complex post call/event properties (per contract, method, or globally) just a few lines of code. It should contextualize all important info about the current state of execution in a test context object.

For more info, see here:
https://github.com/trailofbits/medusa/wiki/API-Overview-(WIP)#extending-testing-methodology

We envision this should look like for generic callback that just broadly take arguments for any call:
`fuzzer.Testing.AddPostCallTest(contractName, functionName, callBackThatProvidesArgsAndTestContext)`
`fizzer.Testing.AddPostEventEmitTest(eventID, callback)`

While we can also make method specific arguments that need to match the signature (but in Go) and will direct outputs into that callback's input, so we can (e.g. match our property test callback with event argument types, and the test suite will populate them for us through reflection maybe)?

Xenomega:
More info to extend this later (in a separate issue that should be opened following this one):

```
A front end testing suite with consideration for Go interface generation for Solidity smart contracts. Building a testing harness around this that plugs into medusa will allow for complex contract deployments, simplification of testing provider code (assertion, property, others), capabilities to write complex custom testing logic with ease, custom cheat code integration, and more. This will inherently tell us where we should invest into API improvements to make querying of complex data easier, and help us standardize our Go API.
```

PR #169, opened 2023-06-26
Added OnlyCallFromDeploymentOrder config option to restrict the contracts to call during a fuzzing campaign

ggrieco-tob:


anishnaik:
As mentioned in the associated issue #132, we will actually merge this capability into `testAllContracts`. @Xenomega will handle this feature while he works on #51.

PR #171, opened 2023-06-27
Custom logger for unstructured and structured logging

### LINKED ISSUE #19: Refactor logging to make use of structured logging providers
Currently the project still has primitive `Printf` logging as a remnant from the project's inception. This should be replaced with a sensible structured logger, which can output console text, or output structure JSON logs to a file. The API should be consistent/clean throughout the application so an external application can parse structured logs and discern events from each other and act upon them sensibly.

It should also be flexible enough that we can choose not to output to stdout/stderr, or output to any arbitrary channel(s). For instance, logging to only JSON files, or logging to CLI and JSON files at the same time.

Other TODOs (thanks @ahpaleus):
- [ ] Quiet and verbose options
- [x] Make UI more colorful/transparent
- [ ] medusa's version is output on startup

anishnaik:
This PR closes #19 and introduces a custom logging API that has the capabilities to support both structured and unstructured logging. The logging API works as follows:
```
logging.Info("Hello", colors.Red, "World", colors.Reset, 1234, StructuredLogInfo{"key": "value"})
```

The API effectively accepts a variadic set of arguments. These arguments are evaluated one by one and are dealt with accordingly:
1. If you are a "color" (e.g. `colors.Red`) then the next `n` arguments are colored with that color until we hit a `colors.Reset`.
2. If you are a `StructuredLogInfo` then that info becomes a set of key-value pairs for that log object.
3. Otherwise, you are just `fmt.Sprintf("%v")`.

The API also allows you to create sub-loggers. Thus, there is one `GlobalLogger` that is instantiated in `NewFuzzer` and each "module" (e.g. fuzzer, corpus) can create a sub-logger that can be used to contextualize each module's log message with their own "module" key-value pair. 

Each `Logger` consists of a `consoleLogger` and a `multiLogger`. The `consoleLogger` is used for console output whereas the `multiLogger` holds a list of `writers` that can perform both structured or unstructured logging to an arbitrary channel. This can be useful for third-party integrations down the line. Currently, at setup the `GlobalLogger` is enabled with console logging and _unstructured_ logging to one file (if requested by the config). Then, the fuzzer creates its own sub-logger off of that. The _downside_ of this approach is that if you `AddWriter` to the `GlobalLogger`, then any sub-loggers branched off of it do not have access to the new writer. This downside should be evaluated during this PR review. Note that due to this design, the `Fatal` log level is not supported. This is because it is not possible to log to both channels before hitting an `os.Exit(1)`. 

We also introduce the idea of a `LogBuffer` in this PR that allows you to create complex, formatted strings for log output (e.g. execution tracer). A `LogBuffer` should ideally be used only right before logging or if a `String()` capability is required. Otherwise, even an `[]any` is sufficient for logging.

The following should be evaluated before merging this PR:

- [x] There are no unit tests for this PR. The compiled binary should be run on a few examples and the console (and file) output should be evaluated. 
- [x] Does the current capability of `AddWriter` / `RemoveWriter` deter any future integrations. Is the use of a `GlobalLogger` the best way to approach this? How do we handle adding new writers to a variety of different sub-loggers?
- [x] Is the color scheme too "colorful"?
- [x] Are there any instances where we need to add more logging? Maybe less?
- [x] Is the API flexible enough to support our various use cases?
- [x] Are there any cases of crappy / clunky code that can be improved?


fuzzing/fuzzer.go
```diff
@@ -664,15 +711,15 @@ func (f *Fuzzer) printMetricsLoop() {
 		secondsSinceLastUpdate := time.Since(lastPrintedTime).Seconds()
 
 		// Print a metrics update
-		fmt.Printf(
-			"fuzz: elapsed: %s, call: %d (%d/sec), seq/s: %d, resets/s: %d, cov: %d\n",
+		f.logger.Info(fmt.Sprintf(
+			"fuzz: elapsed: %s s, calls: %d (%d/sec), seq/s: %d, resets/s: %d, coverage: %d",
```
anishnaik, line 715: Do we want this log object to become more colorful? We can make throw it all into a `StructuredLogInfo` as well and that will add some coloring by defualt.

logging/log_buffer.go
```diff
@@ -0,0 +1,33 @@
...
+		elements: make([]any, 0),
+	}
+}
+
+// Append appends a variadic set of elements to the list of elements
+func (l *LogBuffer) Append(newElements ...any) {
+	l.elements = append(l.elements, newElements...)
+}
+
+// Args returns the list of elements stored in this LogBuffer
```
Xenomega, line 24: This comment addresses the wrong variable name, small update to be made

fuzzing/fuzzer.go
Xenomega: I think it'd be nice :)

Xenomega:
So I'm well through this review. Generally, I think it looks pretty great.

My thoughts so far is:
- I don't love the manual construction of `StructuredLogInfo { "error": <err> }` each time, so it'd be nice to wrap that somehow.
  - In other frameworks, you often can just do `logger.Error(err)`, but I get that isn't great here for coloration. 
  - It'd be nice to use StructuredLogInfo for "additional fields" and in the case of `logger.Error()`, it'd be nice to have it supplied without providing additional fields optionally. Maybe like supporting the following cases (1) log a string, (2) log an err, (3) log a string and an error (and append as you normally would), (4) support structured log info for additional info.
- Otherwise, the only other thing I still need to review, which I can give feedback on later, is the actual JSON output. I don't think we output this based on comments in the `config.go` file, if so, that's fine and we can plan this out later.
  - I think I'd just support the following cases later:
     - Have base fields every log contains (message, level, date/time, and maybe an error field that's nil/empty for non-error messages).
     - Have additional fields populated in a field in our JSON output called "additionalInfo" or "details" or something.
     
Hopefully that makes sense. I think if we can cook up a good solution to the first major bullet point, I'd generally be good to merge here.

fuzzing/config/config.go
```diff
@@ -140,11 +143,38 @@ type OptimizationTestingConfig struct {
 	TestPrefixes []string `json:"testPrefixes"`
 }
 
+// LoggingConfig describes the configuration options for logging to console and file
+type LoggingConfig struct {
+	// Level describes whether logs of certain severity levels (eg info, warning, etc.) will be emitted or discarded.
+	// Increasing level values represent more severe logs
+	Level zerolog.Level `json:"level"`
```
Xenomega, line 150: I think we should just quickly add validation for this field before merging it, for a nicer error. Currently if you run it you'll get a vague error that doesn't tell you it's regarding logging:
```
error Failed to run the fuzz command error="Unknown Level String: 'blah', defaulting to NoLevel"
```

anishnaik: Unfortunately, this occurs during an unmarshaling operation and cannot be caught early. resolving...

logging/log_buffer.go
anishnaik: Resolved in [1c402e4](https://github.com/crytic/medusa/pull/171/commits/1c402e48c7ec898da7c3495d94af79f7e162f5f8)

fuzzing/fuzzer.go
anishnaik: Resolved in [347e76a](https://github.com/crytic/medusa/pull/171/commits/347e76ac1510af28a0345a3e59d34da9a6fb3197)

anishnaik:
The PR has been updated as follows:
1. The API also now provides the ability to pass an `error` directly as part of the variadic list of arguments. This allows for easier use of the API instead of throwing the error into the `StructuredLogInfo` object. Note that only 1 error is supported per log message
2. The `StructuredLogInfo` object is no longer represented as a list of key-value pairs in the log output but instead as a single JSON dictionary. 
e.g.
Before: `{"msg": "message", "key": value", "key2": "value2"}
After: `{"msg": "message", "info": {"key": "value", "key2": "value2"}
3. Errors can be outputted at any log level - so even a trace log can have an error attached to it
4. The "fuzz:..." messages have become colorized

Xenomega (approved):
bing bong

PR #202, opened 2023-08-17
Added denomination parsing to AST value extraction

### LINKED ISSUE #156: Unable to invalidate invariant
In the following contract:
```
// SPDX-License-Identifier: GPL-3.0

pragma solidity ^0.7.6;

/**
 * @title SimpleBank
 * @dev Lets people deposit and withdraw funds
 */
contract SimpleBank {
    mapping(address => uint256) public balances;
    uint256 minimumBalance = 1 ether;

    /**
     * @dev Deposit funds into contract
     */
    function deposit() public payable {
        balances[msg.sender] += msg.value;
        totalDeposited += msg.value;
    }

    /**
     * @dev Withdraw funds from contract
     * @param amount - The amount of ether to withdraw
     */
    function withdraw(uint256 amount) public checkMinimumAndUnderflow(amount) {
        balances[msg.sender] -= amount;

        (bool sent,) = payable(msg.sender).call{value: amount}("");
        require(sent, "Failed to withdraw Ether");
    }

    /**
     * @dev Burns a specified amount of ETH - This will cause the balance of msg.sender
     *      to be deducted but the ETH will still be held by the contract
     * @param amount - The amount to ether to burn
     */
    function burn(uint256 amount) public checkMinimumAndUnderflow(amount) {
        balances[msg.sender] -= amount;
    }

    /**
     * @dev Checks for to ensure that:
     *          - msg.sender has a least the specified minimum balance
     *          - The remaining balance after amount is subtracted from the balance is still over
     *            the minimumBalance
     *          - The specified amount of ether to deduct does not cause an underflow
     * @param amount - The amount expected to deduct from the msg.sender's balance
     */
    modifier checkMinimumAndUnderflow(uint256 amount) {
        // check for minimum balance
        require(balances[msg.sender] > minimumBalance, "Can't remove funds if you have less than the minimum balance");
        require(balances[msg.sender] - amount >= minimumBalance, "Cannot remove funds under minimum balance");

        // check for underflow
        require(amount < balances[msg.sender] + minimumBalance, "Underflow detected! Reverting!");
        _;
    }

    // *** ECHIDNA FUZZING BELOW ***
    uint256 totalDeposited = 0;

    function echidna_test() public returns (bool){
        // echidna test logic goes here
        return totalDeposited >= balances[msg.sender];
    }
}
```
medusa is unable to invalidate the invariant. Echidna however is able to (using seed: -4143723585328282391). I ran about 20 million iterations with medusa before killing it.

The code above has a bug that would allow the `balance[msg.sender]` amount to underflow to a greater number than the total deposited ether in the contract (`totalDeposited`). As an example, user A deposits 1.1 ether and then tries to withdraw 1.5 ether.
The buggy line that allows this is here:
```
        // check for underflow
        require(amount < balances[msg.sender] + minimumBalance, "Underflow detected! Reverting!");
```

Xenomega:
This PR adds denomination parsing to AST value extraction. This means types such as `wei`, `gwei`, `ether`, `weeks`, `years` should now be supported. This PR closes #156 , which was not solveable because the denomination was not parsed, so the correct value was not extracted for the `ValueSet`.

Unrelated changes:
- Added memory usage to printed metrics. Memory usage and worker resets now only show if your log level is DEBUG or lower.

### DIFF
--- a/fuzzing/fuzzer.go
+++ b/fuzzing/fuzzer.go
@@ -12,6 +12,7 @@ import (
 	"math/big"
 	"math/rand"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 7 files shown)


anishnaik (approved):
bing bong

PR #222, opened 2023-09-05
Support NoColor and --no-color

### LINKED ISSUE #217: Support `NoColor` and `--no-color` for terminal output

Exca-DK:
closes #217

### DIFF
--- a/cmd/fuzz_flags.go
+++ b/cmd/fuzz_flags.go
@@ -70,7 +70,7 @@ func addFuzzFlags() error {
 		fmt.Sprintf("print the execution trace for every element in a shrunken call sequence instead of only the last element (unless a config file is provided, default is %t)", defaultConfig.Fuzzing.Testing.TraceAll))
 
 	// Logging color
-	fuzzCmd.Flags().Bool("no-color", false, "disable colored output")
+	fuzzCmd.Flags().Bool("no-color", false, "disabled colored terminal output")
 
 	return nil
 }
--- a/cmd/fuzz_flags.go
+++ b/cmd/fuzz_flags.go
@@ -68,6 +68,10 @@ func addFuzzFlags() error {
 	// Trace all
 	fuzzCmd.Flags().Bool("trace-all", false,
 		fmt.Sprintf("print the execution trace for every element in a shrunken call sequence instead of only the last element (unless a config file is provided, default is %t)", defaultConfig.Fuzzing.Testing.TraceAll))
+
+	// Logging color
+	fuzzCmd.Flags().Bool("no-color", false, "disable colored output")
+
 	return nil
 }
 
@@ -172,6 +176,14 @@ func updateProjectConfigWithFuzzFlags(cmd *cobra.Command, projectConfig *config.
 	// Update trace all enablement
 	if cmd.Flags().Changed("trace-all") {
 		projectConfig.Fuzzing.Testing.TraceAll, err = cmd.Flags().GetBool("trace-all")
+		if err != nil {
+			return err
+		}
+	}
+
+	// Update logging color mode
+	if cmd.Flags().Changed("no-color") {
+		projectConfig.Logging.NoColor, err = cmd.Flags().GetBool("no-color")
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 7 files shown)


cmd/fuzz_flags.go
```diff
@@ -68,6 +68,10 @@ func addFuzzFlags() error {
 	// Trace all
 	fuzzCmd.Flags().Bool("trace-all", false,
 		fmt.Sprintf("print the execution trace for every element in a shrunken call sequence instead of only the last element (unless a config file is provided, default is %t)", defaultConfig.Fuzzing.Testing.TraceAll))
+
+	// Logging color
+	fuzzCmd.Flags().Bool("no-color", false, "disable colored output")
```
anishnaik, line 73: Can we update this to say "disabled colored terminal output"

anishnaik:
@Exca-DK left some comments, mostly around unit testing. Otherwise, PR is looking great. Once those are resolved, I will do one more review and some more testing and then we can merge.

Super super excited for you to be our first external contributor!

Exca-DK:
I don't see any comments :sweat_smile:

logging/colors/colorize_unix.go
```diff
@@ -5,11 +5,21 @@ package colors
...
 
-// EnableColor is a no-op function for non-windows systems because we know that they support ANSI escape codes
-func EnableColor() {}
+var enabled = true
+
+// EnableColor is a simple function for non-windows systems because we know that they support ANSI escape codes
+func EnableColor() { enabled = true }
+
+// DisableColor is a simple function for non-windows systems because we know that they support ANSI escape codes
+func DisableColor() { enabled = false }
```
anishnaik, line 14: Let's update the inline comments for `EnableColor` and `DisableColor` so that they describe the functionality a bit more. I know it's a bit pedantic but want to make sure that we follow best practices on this.

logging/logger_test.go
```diff
@@ -45,3 +50,24 @@ func TestAddAndRemoveWriter(t *testing.T) {
 	assert.Equal(t, len(logger.unstructuredColorWriters), 0)
 	assert.Equal(t, len(logger.structuredWriters), 0)
 }
+
+func TestDisabledColors(t *testing.T) {
```
anishnaik, line 54: Can we add a comment for `TestDisabledColors` describing what it tests?

logging/logger_test.go
```diff
@@ -45,3 +50,24 @@ func TestAddAndRemoveWriter(t *testing.T) {
...
+	// We should expect the underlying data structures are correctly updated
+	assert.Equal(t, len(logger.unstructuredColorWriters), 1)
+
+	// Disable colors and log msg
+	colors.DisableColor()
+	logger.Info("foo")
+
+	// Ensure that msg doesn't include colors afterwards
+	prefix := fmt.Sprintf("%s %s", colors.LEFT_ARROW, "foo")
+	_, ok := strings.CutPrefix(buf.String(), prefix)
```
anishnaik, line 71: So i'm not exactly sure if this tests whether colors are being shown or not? Note that `colors.LEFT_ARROW`, although in the `colors` package is not a color. LEFT_ARROW is a Unicode character. Colors, for terminal output, are actually defined as ANSI escape codes. You can learn more here: https://www.lihaoyi.com/post/BuildyourownCommandLinewithANSIescapecodes.html

So, this test would naturally pass because `logger.Info(foo)` would result in `-> foo` which is equivalent to the prefix. I would recommend changing `logger.Info(foo)` to `logger.Info(colors.Green, "foo", colors.Reset)`. In this way, "foo" should be colored green but since colors are disabled, the prefix should correctly be found without any ANSI escape codes.

Note, that you can test the negative case as well where you re-enable colors and assert that the prefix is _not_ found (aka `ok` is false). This can all be done in the same unit test.

anishnaik (changes requested):

anishnaik:
@Exca-DK apologies - I forgot to submit my review. Please see above for additional comments.

logging/logger_test.go
anishnaik: Okay, agree on the both. Resolving this comment.

anishnaik (approved):
lgtm

anishnaik:
@Exca-DK approved the PR. Feel free to squash and merge whenever you want

PR #255, opened 2023-11-08
Fix `warp` cheatcode

### LINKED ISSUE #235: Timestamp and block number start very low
I'm logging the timestamp and block numbers here:
```
 => [event] Debug0("block.timestamp", 2)
 => [event] Debug0("block.number", 2)
```

This causes a problem for my fuzzing setup because I want to schedule events in the past, which gives me an underflow if I try to calculate them.

I've tried using the `warp` and `roll` cheatcodes in the constructor but got the following error:
```
error Failed to initialize the test chain
‣ contract deployment tx returned a failed status: execution reverted
```

anishnaik:
Our `warp` cheatcode accepted a `uint64` even though the official cheatcode intakes a `uint256`. This PR fixes that.
This PR closes #235

### DIFF
--- a/chain/standard_cheat_code_contract.go
+++ b/chain/standard_cheat_code_contract.go
@@ -17,6 +17,9 @@ import (
 // StandardCheatcodeContractAddress is the address for the standard cheatcode contract
 var StandardCheatcodeContractAddress = common.HexToAddress("0x7109709ECfa91a80626fF3989D68f67F5b1DD12D")
 
+// MaxUint64 holds the max value an uint64 can take
+var _, MaxUint64 = utils.GetIntegerConstraints(false, 64)
+
 // getStandardCheatCodeContract obtains a CheatCodeContract which implements common cheat codes.
 // Returns the precompiled contract, or an error if one occurs.
 func getStandardCheatCodeContract(tracer *cheatCodeTracer) (*CheatCodeContract, error) {
@@ -74,8 +77,7 @@ func getStandardCheatCodeContract(tracer *cheatCodeTracer) (*CheatCodeContract, 
 
 			// Retrieve new timestamp and make sure it is LEQ max value of an uint64
 			newTime := inputs[0].(*big.Int)
-			_, uint64Max := utils.GetIntegerConstraints(false, 64)
-			if newTime.Cmp(uint64Max) > 0 {
+			if newTime.Cmp(MaxUint64) > 0 {
 				return nil, cheatCodeRevertData([]byte("warp: timestamp exceeds max value of type(uint64).max"))
 			}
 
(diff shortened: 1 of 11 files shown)


chain/standard_cheat_code_contract.go
```diff
@@ -67,13 +67,23 @@ func getStandardCheatCodeContract(tracer *cheatCodeTracer) (*CheatCodeContract,
...
+		"warp", abi.Arguments{{Type: typeUint256}}, abi.Arguments{},
 		func(tracer *cheatCodeTracer, inputs []any) ([]any, *cheatCodeRawReturnData) {
 			// Maintain our changes until the transaction exits.
-			original := tracer.evm.Context.Time
-			tracer.evm.Context.Time = inputs[0].(uint64)
+			originalTime := tracer.evm.Context.Time
+
+			// Retrieve new timestamp and make sure it is LEQ max value of an uint64
+			newTime := inputs[0].(*big.Int)
+			_, uint64Max := utils.GetIntegerConstraints(false, 64)
```
Xenomega, line 77: This works, I'm fine with it, but maybe even better would be to make this global, so it doesn't need to be recomputed each time. Should be dirt cheap though, so I'm fine with it.

fuzzing/testdata/contracts/cheat_codes/vm/warp.sol
```diff
@@ -1,6 +1,6 @@
 // This test ensures that the block timestamp can be set with cheat codes
 interface CheatCodes {
-    function warp(uint64) external;
+    function warp(uint256) external;
```
Xenomega, line 3: Question, should we leave compatibility for both uint256 and uint64 in case someone is already dependent on uint64, or do you think it doesn't matter? That's the only thing I thought worth mentioning here.

anishnaik: Given that most people use Foundry, they will already have an interface setup to use with medusa and I doubt they will add the `warp(uint64)` function signature. So, we _can_, but I don't think it will serve any major value since most people will use HEVM's or Foundry's predefined interface.

chain/standard_cheat_code_contract.go
anishnaik: Updated in [8922a67](https://github.com/crytic/medusa/pull/255/commits/8922a6781d101d4486cd688e8dbfe5faa058a21d)

fuzzing/testdata/contracts/cheat_codes/vm/warp.sol
anishnaik: The primary thing we have to do is make sure we highlight this deviation in our documentation, which I have made note of.

Xenomega: Awesome, thanks for the context :)

Xenomega (approved):
bing bong

Issue #248, opened 2023-10-30
Feature Request: add `filterFunctions` capability


anishnaik:
Similar to Echidna, it would be good to have a `filterFunctions` config option that allows you to whitelist specific functions to call

PR #275, opened 2024-01-07
Feat: filter functions

damilolaedwards:
This PR (closes issue #248 ) adds `FilterFunctions` and `FilterBlacklist` options to the fuzzer's configuration. When `FilterBlacklist` is set to true, specified functions in `FilterFunctions` array are excluded from fuzzing; when false, only these functions are included. The FilterFunctions array should consist of contract names and function signatures for blacklisting/whitelisting. Additional validation logic added to ensure the array is not empty when whitelisting is enabled.


0xalpharush:
Right now, it's essentially like we have a default arg of `filterWhitelist: "all"` and `filterBlackList: "none"`. Maybe it would make sense to allow blacklisting "all" so that it doesn't require listing all sig's out. We should also probably warn or even error if a sig is in both lists

Issue #230, opened 2023-09-14
bug: medusa silently fails when provided a CLI flag that is not valid


anishnaik:

0xalpharush:
I am getting `unknown flag: --invalid` already on master so I'm not sure what the issue is. I tried the fix from https://github.com/crytic/medusa/pull/281 in https://github.com/crytic/medusa/tree/fix/raise-cli-errors and get the following instead `Error: unknown flag: --garbage
unknown flag: --garbage`

anishnaik:
Looks it is handled properly.

Issue #192, opened 2023-08-01
TBD: Fail on reverts


anishnaik:
It might make sense, based on discussions with @0xPhaze, to fail on reverts. This would reduce the reliance on always using coverage reports to debug and also enforces that developers are more careful with their preconditions and scoping of testing.

The change would be straightforward since the assertion test case provider can look for Panics as well as EVM reverts and notify the fuzzer on both, instead of just on Panics.

PR #281, opened 2024-01-11
Feat: Fail on revert 

damilolaedwards:
This PR (closes issue #192) implementing fail on reverts when turned on in the config.
Additional fix (closes issue #230) updates the cobra provider setup to throw error when invalid flag is passed in the cli


anishnaik:
Putting a hold on the review for this mainly because it is beneficial to have `vm.expectRevert` to be implemented and test both features together.

PR #312, opened 2024-03-03
Output non-fuzzing failure related error messages

Xenomega:
If you input a bad command/argument, the error is not displayed. This is because `main.go` never does print it. This fixes that issue.

This prints the error if any had occurred. Question for you @anishnaik , do you think this should instead use the global logger? We could do that now or in the future (if it's not a big concern).

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,7 +16,7 @@ func main() {
 	err, exitCode = exitcodes.GetInnerErrorAndExitCode(err)
 
 	// If we have an error, print it.
-	if err != nil {
+	if err != nil && exitCode != exitcodes.ExitCodeFuzzerError {
 		fmt.Println(err)
 	}
 
--- a/main.go
+++ b/main.go
@@ -13,7 +13,7 @@ func main() {
 
 	// If we have a special (non-zero/success) error code, then exit with it.
 	exitCode := exitcodes.GetErrorExitCode(err)
-	if exitCode != 0 {
+	if exitCode != exitcodes.ExitCodeSuccess {
 		fmt.Println(err.Error())
 		os.Exit(exitCode)
 	}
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -10,6 +11,10 @@ func main() {
 	// Run our root CLI command, which contains all underlying command logic and will handle parsing/invocation.
 	err := cmd.Execute()
 
-	// Determine the exit code from any potential error and exit out.
-	os.Exit(exitcodes.GetErrorExitCode(err))
+	// If we have a special (non-zero/success) error code, then exit with it.
+	exitCode := exitcodes.GetErrorExitCode(err)
+	if exitCode != exitcodes.ExitCodeSuccess {
```
anishnaik, line 16: How would this work if there is a failing test case (exit code 7)? Wouldn't that result in getting that error printed as well? I think the issue that Emilio mentioned with medusa's CLI silently exiting when provided a bad flag is separate?

Xenomega: Right no no error bubbled up to main is reflected. So even if an error occurred in cobra, we wouldn't be emitting it.

Xenomega: Also good catch, I can unwrap the inner error to make sure it doesn't exist, so we don't log anything (should cover the case of exit code 7)

anishnaik:
Screenshot 2024-03-27 at 1 56 14 PM

The problem in the above is that a fuzzer error (not a test failure) gets logged and also gets printed out here

anishnaik:
Proposed changed here in the latest [commit](https://github.com/crytic/medusa/pull/312/commits/b83320064fa4c4ee157e042f5d000837f9d0a518) introduces a new error code which is `ErrorCodeFuzzerError`. If this error is encountered, then nothing extra is printed in `main.go`. It's a little bit hacky so feedback would be appreciated.

Screenshot 2024-03-27 at 2 27 41 PM

The image below differentiates between a "fuzzer error" (first run of `./medusa fuzz` in the screenshot) and a "general error" (second run of `./medusa fuzz` in the screenshot).

anishnaik (approved):

PR #318, opened 2024-03-05
Add macos-14 M1 runner to CI

### LINKED ISSUE #317: Add M1 build image support to CI/CD
Our CI/CD currently builds for the following platforms:
- Linux (x86-64)
- mac OS (x86-64)
- Windows (x86-64)

It does not build:
- mac OS (ARM64)

Adding a runner for this that builds M1 images will streamline the release process. It's not necessary for running tests, however.

anishnaik:
The build and test steps have been updated to add `macos-14` to the runner matrix. The naming convention for compression and artifact uploads have changed to support this update.

This closes #317

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,7 @@ jobs:
   test:
     strategy:
       matrix:
-        environment: [ubuntu-latest, macos-latest, macos-14, windows-latest]
+        environment: [ubuntu-latest, macos-12, macos-14, windows-latest]
 
     runs-on: ${{ matrix.environment }}
     timeout-minutes: 20
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     needs: [lint, test]
     strategy:
       matrix:
-        environment: [ubuntu-latest, macos-latest, macos-14, windows-latest]
+        environment: [ubuntu-latest, macos-12, macos-14, windows-latest]
 
     runs-on: ${{ matrix.environment }}
     timeout-minutes: 10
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -110,7 +110,7 @@ jobs:
   test:
     strategy:
       matrix:
-        environment: [ubuntu-latest, macos-latest, windows-latest]
+        environment: [ubuntu-latest, macos-latest, macos-14, windows-latest]
```
elopez, line 113:
```suggestion
        environment: [ubuntu-latest, macos-12, macos-14, windows-latest]
```

We should pin `macos-latest` to either 12 or 13, so that it stays being x86 for the nearby future. Apparently the plan is to change `macos-latest` to mean `macos-14` in April 2024 and we would end up with two arm64 builds 😅 

https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

https://github.com/actions/runner-images?tab=readme-ov-file#available-images

.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     needs: [lint, test]
     strategy:
       matrix:
-        environment: [ubuntu-latest, macos-latest, windows-latest]
+        environment: [ubuntu-latest, macos-latest, macos-14, windows-latest]
```
elopez, line 25:
```suggestion
        environment: [ubuntu-latest, macos-12, macos-14, windows-latest]
```

elopez (commented):
This looks reasonable overall, I only have one concern:

anishnaik:
Closed by #342

Issue #304, opened 2024-02-21
Feature Request: Error reporting when constructor reverts


rappie:
It would be nice to have some descriptive error reporting when the `contructor` function reverts.

Ideally, a full trace like the one that is shown with a broken invariant.

Just printing emitted logs during the constructor would already help tremendously to debug any problems.

Issue #320, opened 2024-03-09
display solidity source of current PC and bytecode if execution reverts during deployment


0xalpharush:
If a contract reverts in the constructor during the initial deployment, it'd be helpful to provide the user more context such as the decoded error or message in the require statement. Ideally the source line that caused it.

Currently we just get:
`‣ contract deployment tx returned a failed status: execution reverted`

PR #337, opened 2024-03-20
feat: add execution trace for failed target contract deployments

0xalpharush:
Closes https://github.com/crytic/medusa/issues/304. This is progress towards https://github.com/crytic/medusa/issues/320 but does not handle the source mapping. 
```
contract Y {
    constructor() public {
        revert();
    }
}

contract X {
    constructor() public {
        new Y();
    }
}
```
Screenshot 2024-03-20 at 4 07 23 PM

### DIFF
--- a/fuzzing/fuzzer.go
+++ b/fuzzing/fuzzer.go
@@ -3,10 +3,6 @@ package fuzzing
 import (
 	"context"
 	"fmt"
-	"github.com/crytic/medusa/fuzzing/coverage"
-	"github.com/crytic/medusa/logging"
-	"github.com/crytic/medusa/logging/colors"
-	"github.com/rs/zerolog"
 	"math/big"
 	"math/rand"
 	"os"
@@ -17,6 +13,11 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/crytic/medusa/fuzzing/coverage"
+	"github.com/crytic/medusa/logging"
+	"github.com/crytic/medusa/logging/colors"
+	"github.com/rs/zerolog"
 
 	"github.com/crytic/medusa/fuzzing/calls"
 	"github.com/crytic/medusa/utils/randomutils"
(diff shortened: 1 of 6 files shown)


fuzzing/fuzzer.go
```diff
@@ -405,7 +406,22 @@ func chainSetupFromCompilations(fuzzer *Fuzzer, testChain *chain.TestChain) erro
...
+					// Replay the execution trace for the failed contract deployment tx
+					err = cse.AttachExecutionTrace(testChain, fuzzer.contractDefinitions)
+					if err != nil {
+						fuzzer.logger.Warn("Failed to attach execution trace to failed contract deployment tx: ", err)
+					}
+					if cse.ExecutionTrace == nil {
+						return fmt.Errorf("contract deployment tx returned a failed status: %v", block.MessageResults[0].ExecutionResult.Err)
+					}
+
+					return fmt.Errorf("contract deployment tx returned a failed status: %v\n %s", block.MessageResults[0].ExecutionResult.Err, cse.ExecutionTrace.Log().String())
```
0xalpharush, line 424: This isn't as pretty as the normal traces, but it at least gives some context on where the revert is.

fuzzing/fuzzer.go
```diff
@@ -405,7 +406,22 @@ func chainSetupFromCompilations(fuzzer *Fuzzer, testChain *chain.TestChain) erro
 
 				// Ensure our transaction succeeded
 				if block.MessageResults[0].Receipt.Status != types.ReceiptStatusSuccessful {
-					return fmt.Errorf("contract deployment tx returned a failed status: %v", block.MessageResults[0].ExecutionResult.Err)
+					// Create a call sequence element to represent the failed contract deployment tx
+					cse := calls.NewCallSequenceElement(nil, msg, 0, 0)
```
0xalpharush, line 410: I may have abused the call sequence here as it seems mostly used for "test" transactions and not the target deployment, but this was straightforward to get the execution traces. I think it is fine to leave target contract as `nil` as it makes sense for a deployment, but would appreciate review

anishnaik:
A few other things were changed in this PR:
1. The chain setup function now returns an execution trace. This can be logged to all channels so that users are aware where a deployment failure occurred.
2. Fixed a bug in the log buffer.

anishnaik (approved):

Issue #333, opened 2024-03-15
Feature: Add ability to deterministically deploy bytecode to specific addresses


anishnaik:
Currently, medusa uses a traditional deployment scheme via `CREATE` where the address of a contract is based on the `deployer` nonce. It would be beneficial to also support deterministic deployments. This would support `deployContracts` and `deployBytecode` that exists in echidna at the moment.

Related:  https://github.com/crytic/medusa/issues/89#issuecomment-1999746421

Issue #39, opened 2022-11-03
Support duplicate contract names


Xenomega:
Right now contract names are expected to be unique by the `fuzzing` package (but is supported by `compilation`). Instead we should come up with some contract ID or something similar to differentiate contracts. It has to factor in source file path. It needs to be human readable and intuitive (or we need an alternate approach), as the config file will have users specifying contract names for deployment order (per #11)

This is very low priority for now and not being considered yet.

0xalpharush:
We also need to support it for predeploys https://github.com/crytic/medusa/pull/353

PR #353, opened 2024-05-01
Support deterministic deployment of contracts

anishnaik:
closes https://github.com/crytic/medusa/issues/333


fuzzing/fuzzer.go
```diff
@@ -348,16 +378,29 @@ func chainSetupFromCompilations(fuzzer *Fuzzer, testChain *chain.TestChain) (err
...
 			// If we found a contract definition that matches this definition by name, try to deploy it
 			if contract.Name() == contractName {
+				// Concatenate constructor arguments, if necessary
 				args := make([]any, 0)
 				if len(contract.CompiledContract().Abi.Constructor.Inputs) > 0 {
+					// If the contract is a predeployed contract, throw an error because they do not accept constructor
+					// args.
+					if _, ok := fuzzer.config.Fuzzing.PredeployedContracts[contractName]; ok {
+						return fmt.Errorf("predeployed contracts cannot accept constructor arguments"), nil
+					}
```
anishnaik, line 403: Technically, we could allow for predeployed contracts to take constructor args. However, currently echidna does not support it so for parity i thought of doing the same. It's easy to allow it though.

chain/vendored/apply_transaction.go
```diff
@@ -67,6 +67,7 @@ func EVMApplyTransaction(msg *Message, config *params.ChainConfig, author *commo
 	receipt.GasUsed = result.UsedGas
 
 	// If the transaction created a contract, store the creation address in the receipt.
+	// TODO: This needs to be fixed
```
anishnaik, line 70: This is the bug...the receipt's contract address is incorrectly set for deterministic deployments. 

We can pass in the `TestChainConfig` so that we get the overrides, but how do you identify whether this txn was for an override address? There is no contract Name or anything. You could re-hash the init bytecode (which is the calldata) and use that. Can't think of another solution.

fuzzing/fuzzer.go
```diff
@@ -324,6 +325,35 @@ func (f *Fuzzer) createTestChain() (*chain.TestChain, error) {
...
+
+		// Throw an error if the contract specified in the config is not found
+		if !found {
+			return nil, fmt.Errorf("%v was specified in the predeployed contracts but was not found in the compilation artifacts", contractName)
+		}
+	}
+
+	// Update the test chain config with the contract address overrides
+	f.config.Fuzzing.TestChainConfig.ContractAddressOverrides = contractAddressOverrides
+
```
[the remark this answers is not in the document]
anishnaik: Feedback on maybe placing this elsewhere is also greatly appreciated.

fuzzing/fuzzer.go
```diff
@@ -324,6 +325,35 @@ func (f *Fuzzer) createTestChain() (*chain.TestChain, error) {
...
 	}
 
+	// Identify which contracts need to be predeployed to a deterministic address by iterating across the mapping
+	contractAddressOverrides := make(map[common.Hash]common.Address, 0)
+	for contractName, addrStr := range f.config.Fuzzing.PredeployedContracts {
+		found := false
+		// Try to find the associated compilation artifact
+		for _, contract := range f.contractDefinitions {
+			if contract.Name() == contractName {
+				// Hash the init bytecode (so that it can be easily identified in the EVM) and map it to the
```
0xalpharush, line 335: Long term, we should probably support duplicate contract names by allowing to match on a key like `File.sol:Contract` and warn if only the contract name is provided in the presence of duplicates

fuzzing/fuzzer.go
```diff
@@ -343,21 +373,34 @@ func chainSetupFromCompilations(fuzzer *Fuzzer, testChain *chain.TestChain) (err
...
-			return fmt.Errorf("missing target contracts (update fuzzing.targetContracts in the project config " +
-				"or use the --target-contracts CLI flag)"), nil
+			return fmt.Errorf("missing target contracts"), nil
 		}
 	}
 
-	// Loop for all contracts to deploy
+	// Concatenate the predeployed contracts and target contracts
+	// Ordering is important here (predeploys _then_ targets) so that you can have the same contract in both lists
+	contractsToDeploy := make([]string, 0)
```
0xalpharush, line 382: This can be statically allocated like `len(fuzzer.config.Fuzzing.PredeployedContracts) + len(fuzzer.config.Fuzzing.TargetContracts)`

fuzzing/fuzzer.go
```diff
@@ -324,6 +325,35 @@ func (f *Fuzzer) createTestChain() (*chain.TestChain, error) {
 		Balance: initBalance,
 	}
 
+	// Identify which contracts need to be predeployed to a deterministic address by iterating across the mapping
+	contractAddressOverrides := make(map[common.Hash]common.Address, 0)
```
0xalpharush, line 329:
```suggestion
	contractAddressOverrides := make(map[common.Hash]common.Address, len(f.config.Fuzzing.PredeployedContracts))
```

fuzzing/fuzzer.go
anishnaik: Yea the problem with the static allocation here is that you can't use the append anymore so you have to create two separate for loops to index into the array and set the different values (for target contracts and predeployed contracts).

fuzzing/fuzzer.go
anishnaik: Agreed, changing.

fuzzing/fuzzer.go
anishnaik: Also, with `PredeployedContracts` being a mapping, it may mildly increase complexity for the static allocation win.

fuzzing/fuzzer.go
anishnaik: Yes, agreed. Looping this into:
https://github.com/crytic/medusa/issues/39

chain/config/config.go
```diff
@@ -13,6 +13,9 @@ type TestChainConfig struct {
 
 	// CheatCodeConfig indicates the configuration for EVM cheat codes to use.
 	CheatCodeConfig CheatCodeConfig `json:"cheatCodes"`
+
+	// ContractAddressOverrides describes contracts that are going to be deployed at deterministic addresses
+	ContractAddressOverrides map[common.Hash]common.Address `json:"contractAddressOverrides,omitempty"`
```
0xalpharush, line 18: It looks like this is always initialized in `fuzzer.go` and shouldn't hit a nil deref but maybe `omitempty` could be left off?

0xalpharush:
Is there any reason we can't insert into the stateDB directly?
https://github.com/ethereum/go-ethereum/blob/8d42e115b1cae4f09fd02b71c06ec9c85f22ad4f/cmd/evm/runner.go#L264-L264

chain/config/config.go
anishnaik: so omitempty was intentional because I don't want this configuration parameter to be shown in the config file (`medusa.json`). This is because the user must update `Fuzzing.PredeployedContracts` which is then parsed and then used to update this variable. Thus, showing both may confuse the end user.

fuzzing/fuzzer_test.go
```diff
@@ -396,6 +396,29 @@ func TestDeploymentsInternalLibrary(t *testing.T) {
...
+			config.Fuzzing.PredeployedContracts = map[string]string{"PredeployContract": "0x1234"}
+		},
+		method: func(f *fuzzerTestContext) {
+			// Start the fuzzer
+			err := f.fuzzer.Start()
+			assert.NoError(t, err)
+
+			// Check for any failed tests and verify coverage was captured
+			assertFailedTestsExpected(f, true)
+			assertCorpusCallSequencesCollected(f, true)
```
0xalpharush, line 417: I would expect this to fail without `TestAllContracts` being enabled. Are `PredeployedContracts` treated differently when they're not in the `TargetContracts` and `TestAllContracts = false`

anishnaik: yes, a predeployed contract is not a target contract. when the contract is deployed on-chain and the event is emitted and captured by the fuzzer worker, it will only add the predeployed contract's methods if `TestAllContracts` is true. 

However, there is an interesting caveat with this PR which is you can predeploy a contract and also have it as part of `TargetContracts`. So, you can have a determinstic deployment of contract `A` which is not a target contract but then also deploy `A` as a target contract for fuzzing. Point being: a function in a predeployed contract is only called if (1) it's also a target contract or (2) `TestAllContracts` is true.

anishnaik: You will notice in the solidity test file for this PR, that I explicitly call `triggerFailure` in the `TestContract`'s function. If that isn't there, there is no way to invoke the `triggerFailure` function. However, if we set `TestAllContracts` to true, then you wouldn't have to explicitly call it. 

P.S. I tried running it with `TestAllContracts` as `true` and removing the explicit invocation of `triggerFailure` and it worked.

0xalpharush (approved):

Issue #168, opened 2023-06-26
Assertion checking in view/pure functions is never reached


ggrieco-tob:
Testing the following contract will never fail in medusa:

```solidity
contract C {
 function invariant() public view { assert(false); }
 function g(uint x) public {}
}
```

using the following command line:

`$ medusa fuzz --target assert_view.sol --assertion-mode`

This is because the view/pure functions are removed. This issue is very related with #55 but the solution for that issue is only a mitigation. Very recommended to integrate slither printer that detect assertions.

PR #363, opened 2024-05-29
feat: call constant methods occasionally and allow assertion testing them

0xalpharush:
closes https://github.com/crytic/medusa/issues/168


0xalpharush:
This will also close https://github.com/crytic/medusa/issues/55 but I'm not sure about it's last requirement
> Ensure the corpus.UpdateCorpusAndCoverageMaps method checks if the last call was to a view method. If it was, do not record the call sequence in the corpus at that step, as it's not a coverage-increasing sequence we'd be interested in recording, as it was not state changing.

This can be left for later work, if desired, but I don't followed why we'd special case the last call.

fuzzing/testdata/contracts/assertions/assert_constant_method.sol
```diff
@@ -0,0 +1,148 @@
...
+        uint256 x2 = FullMath.mulDiv(z, d, y);
+        uint256 y2 = FullMath.mulDiv(z, d, x);
+        assert(x2 >= x);
+        assert(y2 >= y);
+
+        assert(x2 - x < d);
+        assert(y2 - y < d);
+    }
+    fallback() external {}
+}
```
anishnaik, line 148: add a new line here

fuzzing/fuzzer_test.go
```diff
@@ -88,6 +89,7 @@ func TestAssertionMode(t *testing.T) {
 				config.Fuzzing.Testing.AssertionTesting.PanicCodeConfig.FailOnIncorrectStorageAccess = true
 				config.Fuzzing.Testing.AssertionTesting.PanicCodeConfig.FailOnOutOfBoundsArrayAccess = true
 				config.Fuzzing.Testing.AssertionTesting.PanicCodeConfig.FailOnPopEmptyArray = true
+				config.Fuzzing.Testing.AssertionTesting.TestViewMethods = true
```
anishnaik, line 92: I don't see where `assert_constant_method.sol` is being tested in `fuzzer_test.go`? Maybe you planned on adding it to the `filePaths` variable above?

anishnaik (changes requested):
Overall, it looks really good. The only concern is that I don't think your test case is being called at the moment.

Additionally, I'm thinking if we move the `TestViewMethods` config variable outside of the `AssertionTesting` structure and put it in `Testing`. I'm thinking that testing view methods as a concept is more general that assertion testing or property testing. So maybe, we just push it up one level in the hierarchy. what do you think?

fuzzing/fuzzer_test.go
0xalpharush: Good catch! https://github.com/crytic/medusa/pull/363/commits/b8fa7f3dbde64b13c919c2230a77674cada33c27

0xalpharush:
> Additionally, I'm thinking if we move the TestViewMethods config variable outside of the AssertionTesting structure and put it in Testing. I'm thinking that testing view methods as a concept is more general than assertion testing or property testing. So maybe, we just push it up one level in the hierarchy. what do you think?

I understand this config to mean that assertion failures in constant methods will be treated as failure if assertion mode is enabled. For property mode, constant methods are called but only if they have the prefix. I suppose we could have `CallViewMethods` up on level and it enables assertion testing constant methods (if `true`) but also allows completely filtering out all view/pure methods (if `false`). 

We could capture this in an issue and decide this after merging and implementing filtering ( I want to replace https://github.com/crytic/medusa/pull/275 to be compatible with these changes and fix several bugs in its implementation)

anishnaik (approved):

Issue #338, opened 2024-03-21
Inconsistency in the corpus encoded and decoded data


tuturu-tech:
**Version**: 0.1.3, but also present in 0.1.2

The ABI encoded data field does not match the ABI decoded values in the `inputValues` field of the corpus call sequences.

**Uint256:**
```
   "data": "0xd2282dc5ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
   "dataAbiValues": {
    "methodSignature": "setUint256(uint256)",
    "inputValues": [
     "0"
    ]
   },

Correctly decoded value:: type(uint256).max => 115792089237316195423570985008687907853269984665640564039457584007913129639935
```
**Int256:**
```
   "data": "0xa53b1c1e8000000000000000000000000000000000000000000000000000000000000000",
   "dataAbiValues": {
    "methodSignature": "setInt256(int256)",
    "inputValues": [
     "-1809251286499854575800706955788874776224760329015246946758866135024060596214"
    ]
   },
   
Correctly decoded value: type(int256).min => -57896044618658097711785492504343953926634992332820282019728792003956564819968
```

**bytes:**
```
   "data": "0xda359dc80000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000004395d46a88bc0b8281b657614465e8e6393f355639456bd545d93226f8348894f725e214005e87795d8be56055a01a9c5031021b12ef798212c843f5ae41dd73be1cbc0a0000000000000000000000000000000000000000000000000000000000",
   "dataAbiValues": {
    "methodSignature": "setBytes(bytes)",
    "inputValues": [
     ""
    ]
   },
```

Issue #229, opened 2023-09-13
In planning: Improve call sequence shrinking


anishnaik:
It seems like echidna has a different heuristic for call sequence shrinking that might be more optimal than medusa's. There should be research into the differences b/w the two strategies and whether medusa can benefit from employing echidna's strategy more closely.

0xalpharush:
I'm not sure why this isn't 100% given we want to try to shrink the value every single time https://github.com/crytic/medusa/blob/e471c52460ae9a37f5e168a96cbb0fbbe4db33a0/fuzzing/fuzzer.go#L473
Echidna will consistently shrink to *roughly*` setStruct((10, 87489408786382356735958, "\NUL", (80, false)))` but medusa will not (run `./medusa fuzz --target fuzzing/testdata/contracts/value_generation/match_structs_xy.sol` in the root of the repo). Perhaps the string shrinking should have a mutation that returns an empty string "".

aviggiano:
Interestingly, on a very simple test & microbenchmark, it seems like Medusa wins:

https://github.com/foundry-rs/foundry/issues/6683

0xalpharush:
One consideration is that the stateless example in this PR should be able to shrunk to 1 call consistently
https://github.com/crytic/medusa/pull/363

0xalpharush:
Remove reverts when shrinking https://github.com/crytic/echidna/pull/1250

PR #374, opened 2024-06-24
fix: use the shrunken abi values as the msg's data

0xalpharush:
During input shrinking, the msg's data was not being re-encoded so there were duplicate entires in the `test_results` directory that seemed unique based on their `inputValues` field but had the same `data` field.

Closes https://github.com/crytic/medusa/issues/338


0xalpharush:
I think this also explains the strange behavior I was seeing with shrinking where the call sequence appeared to have shrunk but remained unchanged 
https://github.com/crytic/medusa/tree/experiment/shrinking

0xalpharush:
It also seems to have improved on this example https://github.com/crytic/medusa/issues/229#issuecomment-1858260699

anishnaik (approved):

anishnaik:
awesome find, thanks @0xalpharush

PR #378, opened 2024-06-25
feat: include reverted seq's in mutation candidate pool

0xalpharush:


0xalpharush:
I believe you and @ggrieco-tob were discussing this, @aviggiano. Would you like to give it a spin and let use know if you see better results

aviggiano:
I might be wrong, but I remember @anishnaik saying Echidna would not use reverted tx all the time, only very now and then. Shouldn't medusa do the same thing? or maybe have a config parameter %

0xalpharush:
@aviggiano Definitely worth fine-tuning this but I figured it was worth having a branch to try it out sooner than later. When we actually get ready to merge this we will also want to implement this as well https://github.com/crytic/echidna/issues/1264

0xalpharush:
https://github.com/crytic/medusa/pull/456 supersedes this

PR #379, opened 2024-06-25
fix: return potential error created by defer

0xalpharush:


dguido:
This fix is correct but the codebase has changed significantly since June 2024 (10+ commits to fuzzer_worker.go). @0xalpharush - could you verify if this bug still exists in the current code and resubmit if needed?

Issue #375, opened 2024-06-25
add warning to `medusa init` and option to force overwrite instead of overwriting by default


0xalpharush:
`medusa init; medusa init` will overwrite the file without warning, causing the settings to be lost

0xalpharush:
add warning and prompt yes or no to overwrite the file here https://github.com/crytic/medusa/blob/08a6eab7353b831f436ce8a317835a986640fc02/cmd/init.go#L139-L144

PR #381, opened 2024-06-25
fix: prompt user for overwrite if file already exists

to4to:
Fixed Override Behavior with option to force overwrite

```
if _, err = os.Stat(outputPath); err == nil {
		// Prompt user for overwrite confirmation
		fmt.Print("The file already exists. Overwrite? (y/n): ")
		var response string
		if _, err := fmt.Scan(&response); err != nil {
			// Handle the error (e.g., log it, return an error)
			cmdLogger.Error("Failed to scan input", err)
			return err
		}

		if response != "y" && response != "Y" {
			fmt.Println("Operation canceled.")
			return nil
		}

	}
 ```

#375

### CODE REVIEW

**Review by to4to — COMMENTED** [review]
Fixed Override Behavior with option to force overwrite

```
if _, err = os.Stat(outputPath); err == nil {
		// File already exists, prompt user for overwrite
		fmt.Print("The file already exists. Overwrite? (y/n): ")
		var response string
		fmt.Scan(&response)
		if response != "y" && response != "Y" {
			fmt.Println("Operation canceled.")
			return nil
		}
	}

```

#375

to4to:
> @to4to Would you mind running go fmt?

Done

0xalpharush:
Thanks @to4to! This was merged as part of https://github.com/crytic/medusa/pull/444

PR #388, opened 2024-06-27
fix: use read-only lock for event callbacks

0xalpharush:


dguido (approved):
LGTM - This is a correct and valuable fix.

**Assessment:**
- ✅ Legitimate performance improvement using RWMutex for read-heavy event publishing
- ✅ Maintains synchronization guarantees - no new race conditions
- ✅ Removes serialization bottleneck in parallel worker event publishing
- ✅ Matches existing RWMutex pattern used elsewhere in codebase (e.g., non_persistent_cache.go)

This allows all workers to publish events concurrently rather than waiting in line.

Issue #386, opened 2024-06-27
upgrading to go-ethereum 1.14.5 to support Cancun opcodes etc


0xalpharush:
https://github.com/ethereum/go-ethereum/releases/tag/v1.14.5

For the memory leak we had, we no longer need to modify go-ethereum to clear the fastcache
https://github.com/ethereum/go-ethereum/blob/master/triedb/hashdb/database.go#L619-L625

PR #397, opened 2024-07-08
feat: support cancun fork

0xalpharush:
Closes https://github.com/crytic/medusa/issues/386

Upgrades to go-ethereum 1.14.6 to support the Cancun fork, migrates our tracing to the new native tracing, and make our use of the stateDB compatible with the API change.  Previously, we were committing the trie for every tx and so we could just restore to that intermediate state and re-execute the tx. Now, the stateDB expects commits to be tied to a block, so I have taken the approach of re-executing the entire tx sequence from the post-deployment state. I believe removing the calls to `commit` from `PendingBlockAddTx` brings a modest performance increase


chain/test_chain.go
```diff
@@ -148,19 +159,24 @@ func NewTestChain(genesisAlloc core.GenesisAlloc, testChainConfig *config.TestCh
...
-	db := rawdb.NewDatabase(keyValueStore)
+	db := rawdb.NewMemoryDatabase()
+	config := &triedb.Config{
+		HashDB: hashdb.Defaults,
+		// PathDB: pathdb.Defaults,
+	}
+	triedb := triedb.NewDatabase(db, config)
+	// TODO	CleanCacheSize: 256,
+	// this is important in keeping the database performant, so it does not need to fetch repetitively.
+	// https://github.com/ethereum/go-ethereum/issues/30099
```
0xalpharush, line 170: I defaulted to use HashDB for now but I am not sure if we should use the catch given we are not committing to triedb any longer and the comments in https://github.com/ethereum/go-ethereum/issues/30099

chain/test_chain_test.go
```diff
@@ -230,6 +231,7 @@ func TestChainDynamicDeployments(t *testing.T) {
 						// Listen for contract changes
 						deployedContracts := 0
 						chain.Events.ContractDeploymentAddedEventEmitter.Subscribe(func(event ContractDeploymentsAddedEvent) error {
+							fmt.Println("ContractDeploymentAddedEventEmitter", event.Contract.Address.Hex())
```
0xalpharush, line 234:
```suggestion
```

fuzzing/fuzzer_hooks.go
```diff
@@ -27,7 +28,7 @@ type FuzzerHooks struct {
 	ChainSetupFunc TestChainSetupFunc
 
 	// CallSequenceTestFuncs describes a list of functions to be called upon by a FuzzerWorker after every call
-	// in a call sequence.
+	// in a call sequence. These must not
```
0xalpharush, line 31:
```suggestion
	// in a call sequence. These must not commit to state
```

chain/test_chain.go
```diff
@@ -563,27 +580,49 @@ func (t *TestChain) CallContract(msg *core.Message, state *state.StateDB, additi
...
 		NoBaseFee:        true,
 		ConfigExtensions: t.vmConfigExtensions,
 	})
+	// Set our block context and chain config in order for cheatcodes to override what EVM interpreter sees.
+	t.pendingBlockContext = &evm.Context
+	t.pendingBlockChainConfig = evm.ChainConfig()
 
+	tx := utils.MessageToTransaction(msg)
+	if evm.Config.Tracer != nil && evm.Config.Tracer.OnTxStart != nil {
+		evm.Config.Tracer.OnTxStart(evm.GetVMContext(), tx, msg.From)
```
anishnaik, line 598: why are we explicitly calling `OnTxStart`?

0xalpharush: We don't call into the code that would normally invoke this hook, so we have to do it manually https://github.com/ethereum/go-ethereum/blob/f59d013e4064adbb092b323781104c07550a8bff/core/state_processor.go#L110-L117

anishnaik: Part of me feels like we can get rid of the vendored code now, right? I just looked at this function and maybe we can directly call this function and let it deal with all the tracer-related logic.

anishnaik: We still need the `ExecutionResult` though so maybe we do need the vendored code...

anishnaik: I think the `OnTxStart`/`OnTxEnd` logic should be moved to the `vendored` package at least for `PendingBlockAddTx`. For `CallContract` I think it is fine for now. Need to reason through this a bit more.

chain/test_chain_tracer.go
```diff
@@ -1,118 +1,150 @@
...
 	// Call the underlying method for each registered tracer.
+
 	for _, tracer := range t.tracers {
-		tracer.CaptureFault(pc, op, gas, cost, scope, depth, err)
+		if tracer.OnCodeChange != nil {
+			tracer.OnCodeChange(a, prevCodeHash, prev, codeHash, code)
+		}
 	}
+
 }
```
anishnaik, line 148: we probably don't need to subscribe to this right now since we don't have the `OnCodeChange` hook being used anywhere?

fuzzing/fuzzer_test.go
```diff
@@ -250,7 +250,10 @@ func TestCheatCodes(t *testing.T) {
 
 				// enable assertion testing only
 				config.Fuzzing.Testing.PropertyTesting.Enabled = false
+				config.Fuzzing.Testing.OptimizationTesting.Enabled = false
 				config.Fuzzing.Testing.AssertionTesting.Enabled = true
+
+				config.Fuzzing.TestChainConfig.CheatCodeConfig.CheatCodesEnabled = true
```
anishnaik, line 256: I think cheatcodes are enabled by default so you don't need this

chain/standard_cheat_code_contract.go
```diff
@@ -125,23 +129,23 @@ func getStandardCheatCodeContract(tracer *cheatCodeTracer) (*CheatCodeContract,
...
-			// Obtain our spoofed difficulty
+			// // Obtain our spoofed difficulty
+			// TODO: Check chain config here to see if the EVM version is before 'Paris'.
+			// if t.evmContext.ChainConfig.LatestFork(t.evmContext.Time)
+			// // Change difficulty in block context.
+			// originalDifficulty := new(big.Int).Set(tracer.chain.pendingBlockContext.Difficulty.Big())
+			// tracer.chain.pendingBlockContext.Difficulty.Set(spoofedDifficulty)
+			// tracer.CurrentCallFrame().onTopFrameExitRestoreHooks.Push(func() {
+			// 	tracer.chain.pendingBlockContext.Difficulty.Set(originalDifficulty)
+			// })
```
anishnaik, line 140: can we remove this? also, can you update the comment around the difficulty cheatcode's purpose?
Right now it says that it "Sets VM block number".

chain/test_chain_deployments_tracer.go
```diff
@@ -14,19 +19,15 @@ type testChainDeploymentsTracer struct {
...
+	evmContext *tracing.VMContext
 
 	// pendingCallFrames represents per-call-frame data deployment information being captured by the tracer.
 	// This is committed as each call frame succeeds, so that contract deployments which later encountered an error
 	// and reverted are not considered. The index of each element in the array represents its call frame depth.
 	pendingCallFrames []*testChainDeploymentsTracerCallFrame
 
-	// selfDestructDestroysCode indicates whether the SELFDESTRUCT opcode is configured to remove contract code.
-	selfDestructDestroysCode bool
+	nativeTracer *TestChainTracer
```
anishnaik, line 30: just make sure to add a comment about what the native tracer is in all the tracers.

chain/test_chain_tracer.go
```diff
@@ -1,118 +1,150 @@
...
 	}
+	innerTracer := &tracers.Tracer{
+		Hooks: &tracing.Hooks{
+			OnTxStart: tracer.OnTxStart,
+			OnTxEnd:   tracer.OnTxEnd,
+			OnEnter:   tracer.OnEnter,
+			OnExit:    tracer.OnExit,
+			OnOpcode:  tracer.OnOpcode,
+			// OnFault:      tracer.OnFault,
+			OnCodeChange: tracer.OnCodeChange,
```
0xalpharush, line 48: OnCodeChange can also be removed for now

chain/cheat_code_tracer.go
```diff
@@ -98,82 +120,63 @@ func (t *cheatCodeTracer) CurrentCallFrame() *cheatCodeTracerCallFrame {
...
+	// Execute our top frame exit hooks.
+	t.callFrames[0].onTopFrameExitRestoreHooks.Execute(false, true)
 }
 
-// CaptureEnd is called after a call to finalize tracing completes for the top of a call frame, as defined by vm.EVMLogger.
-func (t *cheatCodeTracer) CaptureEnd(output []byte, gasUsed uint64, err error) {
-	// Execute all current call frame exit hooks
-	exitingCallFrame := t.callFrames[t.callDepth]
-	exitingCallFrame.onFrameExitRestoreHooks.Execute(false, true)
-	exitingCallFrame.onTopFrameExitRestoreHooks.Execute(false, true)
```
anishnaik, line 131: why was this line not preserved?

anishnaik: nvm i'm wrong.

anishnaik (approved):
overall, looks great. would recommend taking another pass with the additional changes i made to make sure that original behavior of tracers and chain is preserved

PR #400, opened 2024-07-13
feat: add config to target/exclude func sig. by contract

0xalpharush:
Closes https://github.com/crytic/medusa/issues/248 and replaces https://github.com/crytic/medusa/pull/275

I refactored the detection of assertion, property, and optimization tests to be split into groups upfront rather than iterating over all methods and determining if they should be tested constantly in the main fuzzing loop. This grouping occurs after filtering the available contract methods against the list of target function signatures and excluded function signatures (mutually exclusive). Filtering function signatures **does not** disable calling property and optimization tests which will always be tested unless those testing modes are disabled.


fuzzing/utils/fuzz_method_utils.go
```diff
@@ -32,3 +35,24 @@ func IsPropertyTest(method abi.Method, prefixes []string) bool {
 	}
 	return false
 }
+
+// IsAssertionTest checks whether the method is configured by the attached fuzzer to be a target of assertion testing.
+// Returns true if this target should be tested, false otherwise.
+func isAssertionTest(method abi.Method, includeViewMethods bool) bool {
+	// Only test constant methods (pure/view) if we are configured to.
+	return !method.IsConstant() || includeViewMethods
+}
```
anishnaik, line 44: i am going to get rid of this function because technically, from an API-level, an assertion test is an assertion test iff (1) it's not a property test, (2) it's not an optimization test, (3) it's state changing OR you are testing view methods.

So technically this function does not satisfy (1) or (2). I will push up the changes soon.

fuzzing/utils/fuzz_method_utils.go
```diff
@@ -32,3 +35,24 @@ func IsPropertyTest(method abi.Method, prefixes []string) bool {
...
+
+// BinTestByType sorts a contract's methods by whether they are assertion, property, or optimization tests.
+func BinTestByType(contract *compilationTypes.CompiledContract, testCfg config.TestingConfig) (assertionTests, propertyTests, optimizationTests []abi.Method) {
```
anishnaik, line 47: for utility functions like these, I try to minimize the amount of package dependencies. so i'm going to remove the dependency on the `config` package by explicitly passing in the prefixes, etc.

fuzzing/contracts/contract.go
```diff
@@ -66,3 +115,24 @@ func (c *Contract) CompiledContract() *types.CompiledContract {
...
+// AssertionTestMethods returns the methods that are assertion tests.
+func (c *Contract) AssertionTestMethods() []abi.Method {
+	return c.assertionTestMethods
+}
+
+func (c *Contract) AddTestMethods(assertion, property, optimization []abi.Method) {
+	c.assertionTestMethods = assertion
+	c.propertyTestMethods = property
+	c.optimizationTestMethods = optimization
+}
```
anishnaik, line 138: I am going to expose `AssertionTestMethods` and so on because if you are providing a way to read and write to it, you might as well directly expose it. Additionally, you can then individually set each one instead of being forced to do it via `AddTestMethods.`

fuzzing/contracts/contract.go
```diff
@@ -47,6 +61,41 @@ func NewContract(name string, sourcePath string, compiledContract *types.Compile
...
 }
 
+func containsMethod(methods []string, target string) bool {
+	for _, method := range methods {
+		if method == target {
+			return true
+		}
+	}
+	return false
+}
```
anishnaik, line 71: Getting rid of this function because you can use `slice.Contains[T]`

fuzzing/contracts/contract.go
```diff
@@ -35,6 +38,17 @@ type Contract struct {
...
+	// propertyTestMethods are the methods that are property tests.
+	propertyTestMethods []abi.Method
+
+	// optimizationTestMethods are the methods that are optimization tests.
+	optimizationTestMethods []abi.Method
+
+	// assertionTestMethods are ALL other methods that are not property or optimization tests by default.
+	// If configured, the methods will be targeted or excluded based on the targetFunctionSignatures
+	// and excludedFunctionSignatures, respectively.
+	assertionTestMethods []abi.Method
```
anishnaik, line 51: I am making them all `[]*abi.Method`. Since the `Contract` will sit in the heap somewhere until medusa finishes its campaign, we might as well point to the abi methods within the ABI instead of copying-by-value.

anishnaik: nvm changed my mind lol

anishnaik (approved):

PR #424, opened 2024-08-01
Add new mutation strategy - callSeqGenFuncDuplicateAtRandom

s4nsec:
The function duplicates a call sequence element at index N and places
itat index N+1 if N is not equal to one less the length of the
sequence.Otherwise, the duplicate is placed at index N-1.


0xalpharush:
replaced by https://github.com/crytic/medusa/pull/458

PR #425, opened 2024-08-01
Prompt to overwrite config file

s4nsec:


0xalpharush:
We should resolve the conflicts on https://github.com/crytic/medusa/pull/381

PR #441, opened 2024-08-15
docs/interface-renaming

blablalf:
The goal of this pr is to rename `StdCheats` to `IStdCheats` in the [cheatcodes_overview.md](https://github.com/crytic/medusa/blob/master/docs/src/cheatcodes/cheatcodes_overview.md) documentation file.


0xalpharush:
Hi, we aren't accept minor nits like these to avoid enticing air drop farmers

Issue #376, opened 2024-06-25
add call sequence mutation that copies and repeats a call


0xalpharush:
https://github.com/crytic/medusa/blob/master/fuzzing/fuzzer_worker_sequence_generator.go
https://github.com/crytic/echidna/blob/2c72579e72eb2818d63c60a11180e7359f9984e4/lib/Echidna/Mutator/Corpus.hs#L37
https://github.com/crytic/echidna/blob/2c72579e72eb2818d63c60a11180e7359f9984e4/lib/Echidna/Mutator/Array.hs#L43-L51

We may also want to do the same for solidity array (repeat element in a list)

0xalpharush:
Add a new generator ` callSeqGenFuncDuplicateAtRandom` which duplicates a call seq. elem at index N and inserts it at N+1. Will be helpful to reference existing implementation
https://github.com/crytic/medusa/blob/7b67cebd5b36bdc8005f935c0b9c1522c3a514d3/fuzzing/fuzzer_worker_sequence_generator.go#L356

PR #445, opened 2024-08-15
feat: add expansion call sequence mutation

0xalpharush:
closes https://github.com/crytic/medusa/issues/376 (does not implement mutation for solidity array which could also be added)


0xalpharush:
@ggrieco-tob Does echidna ever generate sequences longer than the limit with this mutation or it's prevented somehow?

0xalpharush:
We need to check after a mutation and truncate the sequence

Issue #373, opened 2024-06-24
`vm.snapshot` cannot be called subsequently after restoring to earlier history (not persisted)


0xalpharush:
The [implementation](https://github.com/crytic/medusa/pull/276) of `vm.snapshot` uses go-ethereum's `revertToSnapshot` which will delete "future" states after reverting e.g. a history of [1,2,3] will become [1] if `revertToSnapshot(1)` is called so it's not possible to call `revertToSnapshot(3)` following. This deviates from how Foundry's implementation works (see https://github.com/foundry-rs/foundry/pull/5487).

PR #455, opened 2024-08-23
snapshot and revertTo bug fix

damilolaedwards:
closes issue #373, [fix](https://github.com/crytic/medusa-geth/commit/f07822f3908a830973d0f6e1424822e2533cd997) applied to medusa-geth. The snapshot and revertTo cheatcodes now use the updated function.


anishnaik:
This issue needs more investigation. The logic on how to persist snapshots in `medusa-geth` does not seem to be sufficient to assert that you can time-travel across snapshots. Currently, the test case for this PR reverts snapshots in reverse order (3, 2, 1) - whereas the correct test for this would revert in (1, 2, 3) order. This test would allow you to ensure that you can snapshot to 1 and then snapshot to 3 and all state changes can be asserted on.

I tried changing the unit test and we are failing to properly handle snapshots/reverts.

PR #456, opened 2024-08-23
feat: combine mutable and immutable call sequences folder

anishnaik:
This PR updates the corpus format of medusa by removing the `call_sequences/mutable` and `call_sequences/immutable` directories. Both directories will be merged into the `call_sequences` directory as part of a legacy migration and the sub-directories will be deleted.

Additionally, mutation testing has changed since _any_ coverage-increasing call sequence will now be used for mutation (similar to echidna).

Replaces https://github.com/crytic/medusa/pull/378

### DIFF
--- a/fuzzing/corpus/corpus.go
+++ b/fuzzing/corpus/corpus.go
@@ -30,13 +30,8 @@ type Corpus struct {
 	// coverageMaps describes the total code coverage known to be achieved across all corpus call sequences.
 	coverageMaps *coverage.CoverageMaps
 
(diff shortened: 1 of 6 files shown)


0xalpharush:
For backwards compatibility with corpuses should we migrate the immutable folder and load into the corpus?

anishnaik:
> For backwards compatibility with corpuses should we migrate the immutable folder and load into the corpus?

When we are reading in the corpus, we can maintain the original code that loaded stuff from `call_sequences/mutable` and `call_sequences/immutable`. Or something like that

0xalpharush (approved):

PR #466, opened 2024-08-28
feat: Reversion metrics

### LINKED ISSUE #431: Reversion Metrics
For each function called by a fuzz worker, record whether the call succeeded or reverted. If it reverted, keep track of the revert reason for metrics.

At the end of a fuzz run, count up the total number of calls/reverts/revert reasons for each function and display them to the user in some way. 

Stretch goal: allow the fuzzer to compare the previous run's reversion metrics to the results of the current run. This would allow detecting big issues in the harness or give dopamine hits for improving the harness.

bsamuels453:
Closes #431 

image


fuzzing/config/config_defaults.go
```diff
@@ -64,6 +64,9 @@ func GetDefaultProjectConfig(platform string) (*ProjectConfig, error) {
 				TraceAll:                     false,
 				TargetFunctionSignatures:     []string{},
 				ExcludeFunctionSignatures:    []string{},
+				ReversionMeasurement: ReversionMeasurementConfig{
+					Enabled: true,
```
0xalpharush, line 68:
```suggestion
					Enabled: false,
```
I'm hesistant to leave this enabled by default. Can you also update the [docs](https://github.com/crytic/medusa/blob/master/docs/src/project_configuration/fuzzing_config.md)?

fuzzing/fuzzer.go
```diff
@@ -803,12 +807,15 @@ func (f *Fuzzer) Start() error {
 		return err
 	}
 
+	// Measure reversion stats across all workers
+	cleanup := f.ReversionStats.StartWorker(f.ctx)
+
```
0xalpharush, line 812:
```suggestion
	cancel := f.ReversionStats.StartWorker(f.ctx)
	defer cancel

```
Is there any reason to call cleanup where it is or it can be deferred?

fuzzing/fuzzer.go
```diff
@@ -827,11 +834,16 @@ func (f *Fuzzer) Start() error {
 		err = fuzzerStoppingErr
 		f.logger.Error("FuzzerStopping event subscriber returned an error", err)
 	}
-
 	// Print our results on exit.
 	f.printExitingResults()
 
+	err = f.ReversionStats.BuildArtifact(f.logger, f.ContractDefinitions(), f.config.Fuzzing.CorpusDirectory)
+	if err != nil {
+		f.logger.Error("Failed to convert reversion metrics to an artifact", err)
```
0xalpharush, line 842:
```suggestion
		f.logger.Error("Failed to generate reversion stats report", err)
```

fuzzing/reversion/revert_report.go
```diff
@@ -0,0 +1,163 @@
...
+			copy(funcSel[:], detail.ID[:4])
+			functionLookup[funcSel] = label
+		}
+	}
+	return functionLookup, errorLookup
+}
+
+// decodeSmuggledSolidityRevertReason is used to decode solidity-inserted reverts, and as a backstop for error selectors
+// that we do not have an ABI for.
+func decodeSmuggledSolidityRevertReason(selector [4]byte) string {
```
0xalpharush, line 136: Let's use https://github.com/crytic/medusa/blob/3a9b0fae7888b1c1972c1178ca5022ebd51f5414/compilation/abiutils/solidity_errors.go#L114

fuzzing/reversion/revert_report.go
```diff
@@ -0,0 +1,163 @@
...
+			Name:          funcName,
+			TotalCalls:    runCount,
+			TotalReverts:  revertCount,
+			RevertPct:     float64(revertCount) / float64(runCount),
+			PrevRevertPct: nil,
+			RevertReasons: []*RevertReasonArtifact{},
+		}
+
+		for errSel, errCount := range s.RevertedCallReasons[funcSel] {
+			errName, ok := errLookup[errSel]
```
0xalpharush, line 97: One issue with this approach versus a tracer is that we don't actually know the codehash of the contract that created the revert data and won't handle collisions correctly. Ideally we could use https://github.com/crytic/medusa/blob/master/compilation/abiutils/solidity_errors.go#L90

fuzzing/reversion/reversion_statistics.go
```diff
@@ -0,0 +1,188 @@
...
+	if err == nil {
+		logger.Info("Revert stats written to file: ", path)
+	}
+	return err
+}
+
+// getRevertReason encodes the error from `result` into an errorSelector. If the error is a solidity-inserted revert,
+// the returnData is decoded to identify the type of revert, and this id is smuggled into the last byte of errorSelector.
+// These smuggled errors can be decoded using decodeSmuggledSolidityRevertReason.
+func getRevertReason(result *core.ExecutionResult) errorSelector {
```
0xalpharush, line 173: Let's use
https://github.com/crytic/medusa/blob/3a9b0fae7888b1c1972c1178ca5022ebd51f5414/fuzzing/executiontracer/execution_trace.go#L218

fuzzing/reversion/reversion_statistics.go
```diff
@@ -0,0 +1,188 @@
...
+	fuzzerTypes "github.com/crytic/medusa/fuzzing/contracts"
+	"github.com/crytic/medusa/logging"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"os"
+	"path/filepath"
+)
+
+type ReversionMeasurer struct {
+	incomingReportsQueue chan *RevertReport
```
0xalpharush, line 20:
```suggestion
	incomingReportsQueue chan *RevertReport
```
I would call this something like TxCallMetrics (it may not have reverted) and change RevertArtifact to RevertReport as artifact is often confusing to other devs IME

fuzzing/fuzzer.go
```diff
@@ -841,6 +853,10 @@ func (f *Fuzzer) Start() error {
 			f.logger.Info("Coverage report saved to file: ", colors.Bold, coverageReportPath, colors.Reset)
 		}
 	}
+	err = f.ReversionStats.WriteReport(f.config.Fuzzing.CorpusDirectory, f.logger)
```
0xalpharush, line 856: Should this check whether it's enabled?

0xalpharush:
Could we pick one phrase like "Revert Statistics" and use it consistently right now there's three or four ways and it'll be easier to search in the codebase for if we use the same terms

fuzzing/reversion/revert_report.go
bsamuels453: I think it's okay that we eat that risk since the current approach gives us the following benefits:
1. Tracking reversion stats using the current method has very little overhead. I ran medusa for an hour with the feature both enabled and disabled and the call count difference was <5%. Using a tracer would probably be a lot more impactful to performance.
2. A function selector collision in the entrypoint contract would get flagged by the compiler, so a theoretical function selector collision could only happen in multi-abi mode which nobody uses anyways.
3. If there's an error selector collision, the developer has bigger problems to worry about because this means their off-chain components will never be able to handle errors properly. 
4. I really don't want to rewrite the code ;)

0xalpharush: I'm fine with leaving it as is, but it's worth documenting. The error that causes the revert in the top level call isn't necessarily defined in that contract as often times the revert is bubbled up from other call frames.

fuzzing/reversion/reversion_statistics.go
bsamuels453: me gusta

fuzzing/fuzzer.go
bsamuels453: negative, I let the ReversionStats object keep track of whether it's supposed to be enabled to improve decoupling:

https://github.com/crytic/medusa/blob/79ccf38be53501bee63f1252bd0ef8141c2ac694/fuzzing/reversion/reversion_statistics.go#L117-L119

fuzzing/reversion/reversion_statistics.go
bsamuels453: this code does use GetSolidityPanicCode further down. we pack panicCode into the last byte of the error selector so we don't have to create a special data type for all the maps/counters to key on

fuzzing/fuzzer.go
bsamuels453: I'll find a way to move the cleanup responsibilities elsewhere. the caller probably shouldn't have to worry about this anyways

bsamuels453:
> Could we pick one phrase like "Revert Statistics" and use it consistently right now there's three or four ways and it'll be easier to search in the codebase for if we use the same terms?

agree, I've refactored this to call the component the "reversion reporter"

fuzzing/reversion/revert_report.go
bsamuels453: I'll add a note to the report header. 

For all we know, after using these reports for another invar engagement, we'll want to figure out the proper source of the revert anyways and end up implementing it as a tracer

bsamuels453:
@0xalpharush I'm satisfied with how the feature is working; this PR is ready for final review.

anishnaik (approved):

PR #468, opened 2024-08-28
feat: allow disabling account checks so contract's can be pranked

0xalpharush:


chain/config/config_defaults.go
```diff
@@ -10,6 +10,7 @@ func DefaultTestChainConfig() (*TestChainConfig, error) {
 			CheatCodesEnabled: true,
 			EnableFFI:         false,
 		},
+		SkipAccountChecks: true,
```
anishnaik, line 13: I think by default this should be `false`. I would rather maintain traditional EVM semantics by default.

0xalpharush: I did it this way bc cheat codes are enabled by default (and I think this makes our prank semantics closer to Foundry). Plus, the error message will probably be perplexing. 

I can change it and maybe catch the error to provide a better message to users

anishnaik: That's a fair point - we can enable it by default.

Issue #451, opened 2024-08-22
Remove all trie computations


0xalpharush:
We don't need a state commitment and it's slow so let's mock it and re-execute transactions as needed (we are already doing this for tracing after the latest go-ethereum upgrade). The only trie that is loaded from the db rn is the genesis and/or post-deployment state AFAIK.

https://github.com/crytic/medusa/blob/b81a9bc7a09baa8f699241fa959e466849b9b453/chain/test_chain.go#L828-L830
https://github.com/crytic/medusa/blob/b81a9bc7a09baa8f699241fa959e466849b9b453/chain/vendored/apply_transaction.go#L64

We should simplify the main fuzz loop to just be changing the block num/timestamp in the block context and not doing all the complex stuff with pending/discarding blocks as if it's a real chain IMO
https://github.com/crytic/medusa/blob/b81a9bc7a09baa8f699241fa959e466849b9b453/chain/test_chain.go#L742-L743

bsamuels453:
consider removing bloom filter calculations (CreateBloom) as well

eats up about 4-5% of the cpu time of EVMApplyTransaction

0xalpharush:
Looking at the CPU report of pprof the most notable improvement we can make is probably memoizing this https://github.com/crytic/medusa/blob/3a9b0fae7888b1c1972c1178ca5022ebd51f5414/fuzzing/coverage/coverage_tracer.go#L164-L168
<img width="1876" alt="Screenshot 2024-08-29 at 12 36 17 PM" src="https://github.com/user-attachments/assets/bf39bed6-64e9-49e4-b7c8-9777fc9d92f2">

PR #471, opened 2024-09-04
(WIP) cache codehash for performance

samalws-tob:
#451

### DIFF
--- a/fuzzing/coverage/coverage_maps.go
+++ b/fuzzing/coverage/coverage_maps.go
@@ -18,6 +18,10 @@ type CoverageMaps struct {
 
 	// TODO comment this
 	// The assumption here is that geth codehash matches if and only if codehash matches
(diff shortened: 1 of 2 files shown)


samalws-tob:
Ran benchmark, this removes ExtractContractMetadata from the flamegraph, just about doubles gas/second, and increases calls/second by about 50%

Need to clean this up a lot, I think we can remove the cachedGethCodeHash entirely and keep only one of the gethCodeHashToCodeHash's but I'm not sure which of them to keep, should be easy enough to figure out
Also I need to check the assumption that geth's codehash value is different during init vs during runtime. If it's not true we need to change gethCodeHashToCodeHash to a `map[bool]map[common.Hash]*common.Hash`. Might want to do this anyway just to be safe

samalws-tob:
#472

PR #522, opened 2024-12-16
Add Docker container

elopez:
This adds a Docker container including medusa and crytic-compile. This container should be suitable for running fuzzing jobs:

https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/running-jobs-in-a-container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,53 @@
+## medusa build process
+FROM golang:1.22 AS medusa
+
+WORKDIR /src
+COPY . /src/medusa/
+RUN cd medusa && \
+    go build -trimpath -o=/usr/local/bin/medusa -ldflags="-s -w" && \
+    chmod 755 /usr/local/bin/medusa
+
+
+## Python dependencies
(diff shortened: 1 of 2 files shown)


anishnaik:
@elopez maybe we want to download slither now instead since we have integrated it?

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-suggests --no-install-recommends \
+        gcc \
+        python3 \
+        python3-dev \
+        python3-venv
+ENV PIP_DISABLE_PIP_VERSION_CHECK=1
+ENV PIP_NO_CACHE_DIR=1
+RUN python3 -m venv /venv && /venv/bin/pip3 install --no-cache --upgrade setuptools pip
+RUN /venv/bin/pip3 install --no-cache slither-analyzer solc-select
```
elopez, line 22: @anishnaik it already does :)

anishnaik (approved):

PR #544, opened 2025-01-17
Simplify flake

bohendo:
Now that [crytic.nix](https://github.com/crytic/crytic.nix) is public, we don't need to define nix config for solc-select, crytic-compile, etc in this repo. Instead, we can use the packages provided by crytic.nix.


flake.nix
```diff
@@ -3,122 +3,64 @@
...
-          default = medusa;
+  outputs = inputs: with inputs;
+    utils.lib.eachDefaultSystem (system: let
+      pkgs = import nixpkgs { inherit system; config.allowUnfree = true; };
+    in rec {
 
+      packages = {
+        medusa = pkgs.buildGoModule {
+          pname = "medusa";
+          version = "0.1.8"; # from cmd/root.go
```
anishnaik, line 18: Does this need to be updated every time we make a new release?

bohendo: Nothing will break if it's not updated, `medusa --version` would print the right thing, but it could cause confusion eg:
```
$ which medusa
/nix/store/k1clqlydpg8ajg1srsi1ck6a9cl0mw00-medusa-0.1.8/bin/medusa
```
Ideally, the nix config would pull this value out of the source file where it's defined. I can set that up in the next couple days

anishnaik: So it's not make or break if it cannot be automated, I think it would be better if it is so that we don't have to cut a new release if we screw it up. 

If we can't, then that's fine, and I can add it to my release checklist.

bohendo:
Sprinkled a bit of regex into flake.nix to dynamically read the version from cmd/root.go, a couple of implications:
- This assumes that the version constant defined in cmd/root.go is and remains the source of truth.
- it hardcodes the regex: `.*const version = "([^"]+)".*`, this would need to change if eg we rename this variable, switch to single quote raw strings, etc

anishnaik (approved):

PR #569, opened 2025-02-06
Fine tune coverage

### LINKED ISSUE #392: move the TestViewMethods config variable outside of the AssertionTesting structure
> > Additionally, I'm thinking if we move the TestViewMethods config variable outside of the AssertionTesting structure and put it in Testing. I'm thinking that testing view methods as a concept is more general than assertion testing or property testing. So maybe, we just push it up one level in the hierarchy. what do you think?
> 
> I understand this config to mean that assertion failures in constant methods will be treated as failure if assertion mode is enabled, but I see your point that this will affect the sequences generated for property mode as well. I suppose we could have `CallViewMethods` up one level and it enables assertion testing constant methods (if `true`) but also allows completely filtering out all view/pure methods (if `false`) in either mode.
> 
> We could capture this in an issue and decide this after merging and implementing filtering ( I want to replace #275 to be compatible with these changes and fix several bugs in its implementation)

Yeah I think this is fine for now and we can figure this out later

_Originally posted by @anishnaik in https://github.com/crytic/medusa/issues/363#issuecomment-2203544087_

anishnaik:
Closes #392 

This PR fine tunes coverage by ensuring that view methods are called by default.

### DIFF
--- a/docs/src/project_configuration/testing_config.md
+++ b/docs/src/project_configuration/testing_config.md
@@ -53,7 +53,7 @@ contract MyContract {
 
 - **Type**: Boolean
 - **Description**: Whether `pure` / `view` functions should be called and tested.
-- **Default**: `false`
+- **Default**: `true`
 -
 
 ### `testAllContracts`
--- a/docs/src/project_configuration/testing_config.md
+++ b/docs/src/project_configuration/testing_config.md
@@ -54,7 +54,8 @@ contract MyContract {
 - **Type**: Boolean
 - **Description**: Whether `pure` / `view` functions should be called and tested.
 - **Default**: `false`
-- 
+-
+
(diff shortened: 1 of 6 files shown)


docs/src/project_configuration/testing_config.md
```diff
@@ -49,6 +49,13 @@ contract MyContract {
   the [`timeout`](./fuzzing_config.md#timeout) is hit, or the user manually stops execution.
 - **Default**: `true`
 
+### `testViewMethods`
+
+- **Type**: Boolean
+- **Description**: Whether `pure` / `view` functions should be called and tested.
+- **Default**: `false`
```
elopez, line 56:
```suggestion
- **Default**: `true`
```

Issue #306, opened 2024-02-22
Option to set Medusa initial balance using integers same as Echidna


devdacian:
Echidna allows the initial balance to be set using integers, for example to set the initial balance to type(uint256).max: 

`balanceContract: 115792089237316195423570985008687907853269984665640564039457584007913129639935`

Ideally Medusa would also allow initial balance to be set the same way as this is much more natural than using hex notation.

When using [chisel](https://book.getfoundry.sh/chisel/) to get the hex for various balances then using that hex as input to Medusa, this can result in some weird errors, eg: type(uint256).max 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff

error Failed to initialize the test chain
‣ test chain state write error when adding tx to pending block: insufficient funds for gas * price + value: address 0x0000000000000000000000000000000000030000 have 28948022309329048855892746252171976963317496166410141009864396001978282409983 want 115792089237316195423570985008687907853269984665640564039457584007913129639935

0xalpharush:
I believe the balances can be a base-10 integer. The error seems unrelated to supporting integers and caused by an insufficient balance. Could you share an example of what you're trying to accomplish so we could get clarity on what needs to be fixed?

anishnaik:
@0xZRA I think this can be a good first issue for you. If you would like to work on this let me know and I can provide some pointers on how to get started.

0xZRA:
@anishnaik Happy to look into it and would certainly value your advice on how to get started.

0xZRA:
> hexutils.Big

Thanks @anishnaik, it makes sense overall but just to confirm, the goal is:
- maintain the current hex notation support while adding base-10 Strings as an additional option for setting initial balances;
- document the change in https://github.com/crytic/medusa/blob/master/docs/src/project_configuration/fuzzing_config.md?plain=1#L90-L99
- add test -  I haven't found tests covering config properties, do you recommend adding an entirely new one or expand on existing that may be a good fit?

anishnaik:
So we actually don't want to support hex notation anymore. We want to support only base-10 strings. no need to add a new test, I would just update any tests in `fuzzer_test.go` that use target contract balances if necessary.

0xZRA:
[...]

PR #580, opened 2025-02-14
Support multiple balance formats in targetContractsBalances

0xZRA:
This is to address issue https://github.com/crytic/medusa/issues/306
Relies on running `go generate` command to  automatically generate code generate JSON marshaling code from template for `FuzzingConfig` 

@anishnaik  

I'm running into an issue where tests are failing locally on my machine, even on the `master` branch. I wanted to flag this early - I'm using Sequoia version 15.1.1, so it's unclear if my changes are implemented correctly.

`go test -v ./fuzzing`

```
FAIL
FAIL    github.com/crytic/medusa/fuzzing        240.868s
FAIL
```

At the same time, when running following tests individually thru VSCode, both are passing: 

```
Running tool: /opt/homebrew/bin/go test -timeout 30s -run ^TestDeploymentsWithPayableConstructors$ github.com/crytic/medusa/fuzzing

ok  	github.com/crytic/medusa/fuzzing	6.693s
```

```
Running tool: /opt/homebrew/bin/go test -timeout 30s -run ^TestDeploymentsWithPredeploy$ github.com/crytic/medusa/fuzzing

ok  	github.com/crytic/medusa/fuzzing	1.280s
```


fuzzing/fuzzer.go
```diff
@@ -505,7 +505,10 @@ func chainSetupFromCompilations(fuzzer *Fuzzer, testChain *chain.TestChain) (*ex
 				// If our project config has a non-zero balance for this target contract, retrieve it
 				contractBalance := big.NewInt(0)
 				if len(balances) > i {
-					contractBalance = new(big.Int).Set(balances[i])
+					_, success := contractBalance.SetString(balances[i], 10)
+					if !success {
+						contractBalance = big.NewInt(0)
+					}
```
anishnaik, line 511: You won't need this logic once the change in config.go is made

fuzzing/config/config.go
```diff
@@ -116,7 +115,7 @@ type FuzzingConfig struct {
 // have its custom marshaling methods auto-generated and will handle type conversions for serialization purposes.
 // For example, this enables serialization of big.Int but specifying a different field type to control serialization.
 type fuzzingConfigMarshaling struct {
-	TargetContractsBalances []*hexutil.Big
+	TargetContractsBalances []string
```
anishnaik, line 118: So technically the only change you have to make for this PR is this (and the update to the documentation). Everything else should click into place once `go generate` is run in the `config/` directory.

anishnaik:
@0xZRA provided some feedback. Lmk if you have any questions. I'll run the CI when you make those changes and I can investigate if/why the tests are failing.

fuzzing/config/config.go
0xZRA: @anishnaik I'm getting the following error when running `go generate` from `config` dir:
```
rebase@MacBookPro config % go generate                                                              
-: invalid field override: slice element type string is not convertible to *math/big.Int
exit status 1
config.go:22: running "go": exit status 1
```

anishnaik:
Hey @0xZRA okay so I think we need to change our approach. Tbh I didn't expect `go generate` to complain about this (sorry about that). So, going back to the original problem: we need to (de)serialize the big.Ints. Another way to approach this is to create a custom type (`type ContractBalance *big.Int`). Thus, the `TargetContractBalances` will become of type `[]ContractBalance`. Now, we can write our own custom `MarshalJSON` and `UnmarshalJSON` that will marshal the `ContractBalance` to a string and unmarshal from a string to a `*big.Int`. Here is a stack overflow post that should get you in the right direction: https://stackoverflow.com/questions/53991835/how-to-marshal-and-unmarshal-big-int-in-json

I would like to take it one more step than what is mentioned in the stack overflow post. I would like to be able to unmarshal any three of these string possibilities:
1. **Base-10 string**: "1234" or "45678"
2. **Hex string**: "0xabcdef"
3. **Scientific notation for base-10 strings** (optional if it is easy): "1.2e18"

Let me know if you have any questions.

Also, about the failing tests: We have some challenging unit tests that sometimes takes medusa a while to solve. That's why sometimes the tests may fail to run. Usually, re-running it does the trick but I do need to investigate why it is happening more frequently nowadays.

0xZRA:
> Hey @0xZRA okay so I think we need to change our approach. Tbh I didn't expect `go generate` to complain about this (sorry about that). So, going back to the original problem: we need to (de)serialize the big.Ints. Another way to approach this is to create a custom type (`type ContractBalance *big.Int`). Thus, the `TargetContractBalances` will become of type `[]ContractBalance`. Now, we can write our own custom `MarshalJSON` and `UnmarshalJSON` that will marshal the `ContractBalance` to a string and unmarshal from a string to a `*big.Int`. Here is a stack overflow post that should get you in the right direction: https://stackoverflow.com/questions/53991835/how-to-marshal-and-unmarshal-big-int-in-json
> 
> Doing it this way will remove any requirement for `go generate` so you can remove the directive and the associated file as well.
> 
> I would like to take it one more step than what is mentioned in the stack overflow post. I would like to be able to unmarshal any three of these string possibilities:
> 
> 1. **Base-10 string**: "1234" or "45678"
> 2. **Hex string**: "0xabcdef"
> 3. **Scientific notation for base-10 strings** (optional if it is easy): "1.2e18"
> 
> Let me know if you have any questions.
> 
> Also, about the failing tests: We have some challenging unit tests that sometimes takes medusa a while to solve. That's why sometimes the tests may fail to run. Usually, re-running it does the trick but I do need to investigate why it is happening more frequently nowadays.

@anishnaik  I appreciate your response and I think it makes sense so far. Just wanted to clarify one point in your approach regarding the unmarshalling of the 3 string possibilities. Does the proposed approach still involve maintaining the current hexadecimal notation for setting initial balances, while adding the option to use base-10 strings, including scientific notation?

So if we take it straight to the docs, the new version will look like:
### `targetContractsBalances`

- Type: [Base-16 Strings] (e.g. `[0x123, 0x456, 0x789]`)
- Type: [Base-10 Strings] (e.g. `["123", "1000000", "115792089237316195423570985008687907853269984665640564039457584007913129639935"]`)
- Type: [Base-10 Strings Scientific Notation] (e.g. `["1.23e2", "1.0e6", "1.15792089237316195423570985008687907853269984665640564039457584007913129639935e77"]`)

anishnaik:
Technically, you can also mix and match: `["1e12", "0x1234", "5678"]`. Since you are evaluating each string individually, and not the whole array, the logic shouldn't change depending on what the array looks like. 

Like the `UnmarshalJSON` function you write will take a `[]byte` and you need to convert it into a `ContractBalance` (aka *big.Int). This `[]byte` could be any of the three string possibilities (or invalid).

0xZRA:
> Technically, you can also mix and match: `["1e12", "0x1234", "5678"]`. Since you are evaluating each string individually, and not the whole array, the logic shouldn't change depending on what the array looks like.
> 
> Like the `UnmarshalJSON` function you write will take a `[]byte` and you need to convert it into a `ContractBalance` (aka *big.Int). This `[]byte` could be any of the three string possibilities (or invalid).
> 
> Also, for `MarshalJSON`, always just serialize it as a base-10 string.
> 
> This PR became more complicated than I expected :)

I hadn't thought of mix and match, you're right. I'll take a shot at it.

0xZRA:
This is still WIP

fuzzing/fuzzer_test.go
```diff
@@ -461,14 +461,14 @@ func TestDeploymentsInternalLibrary(t *testing.T) {
...
-			config.Fuzzing.PredeployedContracts = map[string]string{"PredeployContract": "0x1234"}
-			config.Slither.UseSlither = false
+		configUpdates: func(pkgConfig *config.ProjectConfig) {
+			pkgConfig.Fuzzing.TargetContracts = []string{"TestContract"}
+			pkgConfig.Fuzzing.TargetContractsBalances = []config.ContractBalance{{Int: *big.NewInt(1)}}
+			pkgConfig.Fuzzing.TestLimit = 1000 // this test should expose a failure immediately
+			pkgConfig.Fuzzing.Testing.PropertyTesting.Enabled = false
+			pkgConfig.Fuzzing.Testing.OptimizationTesting.Enabled = false
+			pkgConfig.Fuzzing.PredeployedContracts = map[string]string{"PredeployContract": "0x1234"}
+			pkgConfig.Slither.UseSlither = false
```
anishnaik, line 471: Can we revert the change from `pkgConfig` back to `config`? For the sake of uniformity across all tests.

fuzzing/fuzzer_test.go
```diff
@@ -486,13 +486,16 @@ func TestDeploymentsWithPredeploy(t *testing.T) {
...
-			config.Fuzzing.TestLimit = 1 // this should happen immediately
-			config.Fuzzing.Testing.AssertionTesting.Enabled = false
-			config.Fuzzing.Testing.OptimizationTesting.Enabled = false
-			config.Slither.UseSlither = false
+		configUpdates: func(pkgConfig *config.ProjectConfig) {
+			pkgConfig.Fuzzing.TargetContracts = []string{"FirstContract", "SecondContract"}
+			pkgConfig.Fuzzing.TargetContractsBalances = []config.ContractBalance{
+				{Int: *big.NewInt(0)},
+				{Int: *big.NewInt(1e18)},
+			}
```
anishnaik, line 494: I think this is great. Let's add one more for a hex string? So update the fuzz test ("testdata/contracts/deployments/deploy_payable_constructors.sol") to have a `ThirdContract` and send it funds using a hex string.

fuzzing/config/config.go
```diff
@@ -112,11 +109,47 @@ type FuzzingConfig struct {
...
-type fuzzingConfigMarshaling struct {
-	TargetContractsBalances []*hexutil.Big
+type ContractBalance struct {
+	big.Int
+}
+
+func (cb *ContractBalance) UnmarshalJSON(data []byte) error {
+	s := strings.Trim(string(data), "\"")
+
+	if s == "null" || s == "" {
```
anishnaik, line 119: No need to handle the "null" case.

fuzzing/config/config.go
```diff
@@ -112,11 +109,47 @@ type FuzzingConfig struct {
...
+		cb.Int.SetInt64(0)
+		return nil
+	}
+
+	// hex handling
+	if strings.HasPrefix(strings.ToLower(s), "0x") {
+		if _, ok := cb.Int.SetString(s[2:], 16); ok {
+			return nil
+		}
+		return fmt.Errorf("Error: innvalid hex string: %s", s)
```
anishnaik, line 129: Generally in golang errors are all lowercase. So I would recommend making the error a simple 
```
fmt.Errorf("invalid hex string provided while unmarshaling contract balance: %s", s)
```

fuzzing/config/config.go
```diff
@@ -112,11 +109,47 @@ type FuzzingConfig struct {
...
+	}
+
+	// scientific notation handling
+	if strings.ContainsAny(s, "eE") {
+		f, err := strconv.ParseFloat(s, 64)
+		if err == nil {
+			plainStr := strconv.FormatFloat(f, 'f', 0, 64)
+			if _, ok := cb.Int.SetString(plainStr, 10); ok {
+				return nil
```
anishnaik, line 138: We should throw an error if `ok` is false. Similar to the error I mentioned above.

fuzzing/config/config.go
```diff
@@ -112,11 +109,47 @@ type FuzzingConfig struct {
...
+			}
+		}
+	}
+
+	// base-10 string
+	if _, ok := cb.Int.SetString(s, 10); ok {
+		return nil
+	}
```
anishnaik, line 146: Same as above, make sure to handle the error.

fuzzing/config/config.go
```diff
@@ -112,11 +109,47 @@ type FuzzingConfig struct {
...
+
+	return fmt.Errorf("invalid number format: %s", s)
```
anishnaik, line 148: i would highlight, as mentioned above, that this occurred while unmarshaling a contract balance. it provides the user more context.

docs/src/project_configuration/fuzzing_config.md
```diff
@@ -87,15 +87,13 @@ The fuzzing configuration defines the parameters for the fuzzing campaign.
...
 
-- **Type**: [Base-16 Strings] (e.g. `[0x123, 0x456, 0x789]`)
+- **Type**: [Base-10 Strings, Hexadecimal Strings, Scientific notation for base-10 values] (e.g. `["1234", "0x1234", "1.2e18"]`)
 - **Description**: The starting balance for each contract in `targetContracts`. If the `constructor` for a target contract
   is marked `payable`, this configuration option can be used to send ether during contract deployment. Note that this array
-  has a one-to-one mapping to `targetContracts`. Thus, if `targetContracts` is `[A, B, C]` and `targetContractsBalances` is
-  `["0", "0xff", "0"]`, then `B` will have a starting balance of 255 wei and `A` and `C` will have zero wei. Note that the wei-value
-  has to be hex-encoded and _cannot_ have leading zeros. For an improved user-experience, the balances may be encoded as base-10
-  format strings in the future.
+  has a one-to-one mapping to `targetContracts`. Thus, if targetContracts is [A, B, C] and targetContractsBalances is ["1234", "0x1234", "1.2e18"],
```
anishnaik, line 95: make sure to add the "`" for `targetContracts`, `targetContractBalances`, `[A, B, C]`, and ` ["1234", "0x1234", "1.2e18"]`

docs/src/project_configuration/fuzzing_config.md
```diff
@@ -87,15 +87,13 @@ The fuzzing configuration defines the parameters for the fuzzing campaign.
...
-- **Type**: [Base-16 Strings] (e.g. `[0x123, 0x456, 0x789]`)
+- **Type**: [Base-10 Strings, Hexadecimal Strings, Scientific notation for base-10 values] (e.g. `["1234", "0x1234", "1.2e18"]`)
 - **Description**: The starting balance for each contract in `targetContracts`. If the `constructor` for a target contract
   is marked `payable`, this configuration option can be used to send ether during contract deployment. Note that this array
-  has a one-to-one mapping to `targetContracts`. Thus, if `targetContracts` is `[A, B, C]` and `targetContractsBalances` is
-  `["0", "0xff", "0"]`, then `B` will have a starting balance of 255 wei and `A` and `C` will have zero wei. Note that the wei-value
-  has to be hex-encoded and _cannot_ have leading zeros. For an improved user-experience, the balances may be encoded as base-10
-  format strings in the future.
+  has a one-to-one mapping to `targetContracts`. Thus, if targetContracts is [A, B, C] and targetContractsBalances is ["1234", "0x1234", "1.2e18"], 
+  then A will have a starting balance of 1,234 wei, B will have 4,660 wei (0x1234 in decimal), and C will have 1.2 ETH (1.2 × 10^18 wei).
```
anishnaik, line 96: Use the "`" here as well for the contract names

anishnaik (changes requested):
Hey @0xZRA, great job! Provided some feedback. lmk if you have any questions.

fuzzing/config/config.go
```diff
@@ -112,11 +109,47 @@ type FuzzingConfig struct {
...
+}
+
+func (cb ContractBalance) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("\"%s\"", cb.Int.String())), nil
```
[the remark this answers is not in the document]
0xZRA: Doesn't the string need to be quoted, e.g. `"123456"?`

fuzzing/fuzzer_test.go
0xZRA: The renaming was just to reference `config.ContractBalance` without it clashing with the imported `*config.ProjectConfig`. Let me know if declaring `ContractBalance` on the same level `ProjectConfig` is suboptimal and needs to be changed.

fuzzing/fuzzer.go
```diff
@@ -505,7 +506,7 @@ func chainSetupFromCompilations(fuzzer *Fuzzer, testChain *chain.TestChain) (*ex
 				// If our project config has a non-zero balance for this target contract, retrieve it
 				contractBalance := big.NewInt(0)
 				if len(balances) > i {
-					contractBalance = new(big.Int).Set(balances[i])
+					contractBalance = new(big.Int).Set(&balances[i].Int) // FIX
```
anishnaik, line 509: What is the FIX that is needed here?

0xZRA: Not needed, oversight on my side. Will remove, sorry!

fuzzing/fuzzer_test.go
anishnaik: Yup, that makes sense. We can leave it as is.

fuzzing/config/config.go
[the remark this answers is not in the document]
anishnaik: Since you are converting the big.Int to a string via `cb.Int.String()`, you should be fine without having to explicitly add the quotation marks.

fuzzing/config/config.go
```diff
@@ -112,11 +109,53 @@ type FuzzingConfig struct {
...
-	TargetContractsBalances []*hexutil.Big
+type ContractBalance struct {
+	big.Int
+}
+
+// UnmarshalJSON parses JSON data into big.Int from empty strings, hex ("0x"),
+// scientific notation (e/E), and base-10 formats
+
+func (cb *ContractBalance) UnmarshalJSON(data []byte) error {
+	s := strings.Trim(string(data), "\"")
```
anishnaik, line 120: If you remove the quotation marks from the unmarshal, you can remove the trimming here.

fuzzing/config/config.go
```diff
@@ -112,11 +109,53 @@ type FuzzingConfig struct {
...
-// For example, this enables serialization of big.Int but specifying a different field type to control serialization.
-type fuzzingConfigMarshaling struct {
-	TargetContractsBalances []*hexutil.Big
+type ContractBalance struct {
+	big.Int
+}
+
+// UnmarshalJSON parses JSON data into big.Int from empty strings, hex ("0x"),
+// scientific notation (e/E), and base-10 formats
+
```
anishnaik, line 118: Remove this extra new line

fuzzing/config/config.go
```diff
@@ -112,11 +109,53 @@ type FuzzingConfig struct {
...
+		f, err := strconv.ParseFloat(s, 64)
+		if err == nil {
+			plainStr := strconv.FormatFloat(f, 'f', 0, 64)
+			if _, ok := cb.Int.SetString(plainStr, 10); !ok {
+				return fmt.Errorf("invalid format for contract balance (scientific notation) while unmarshaling contract balance: %s", s)
+			}
+			return nil
+		} else {
+			return fmt.Errorf("error parsing scientific notation while unmarshaling contract balance: %w", err)
+		}
```
anishnaik, line 147: This is another nitpick but generally you want to handle the `err != nil` case first instead of the success case. This reduces the nesting structure of this logic. Since if `err != nil`, then you will return an error and then you don't really need an else branch. This sort of "guard clause" structure reduces the amount of indentation and makes the code cleaner.

fuzzing/config/config.go
```diff
@@ -112,11 +109,53 @@ type FuzzingConfig struct {
...
+	}
+
+	// Scientific notation handling
+	if strings.ContainsAny(s, "eE") {
+		f, err := strconv.ParseFloat(s, 64)
+		if err == nil {
+			plainStr := strconv.FormatFloat(f, 'f', 0, 64)
+			if _, ok := cb.Int.SetString(plainStr, 10); !ok {
+				return fmt.Errorf("invalid format for contract balance (scientific notation) while unmarshaling contract balance: %s", s)
+			}
```
anishnaik, line 143: This is a good example of handling the "bad case first". No feedback here, just want to show an example for the other review comment i made for this area.

fuzzing/config/config.go
```diff
@@ -77,7 +74,7 @@ type FuzzingConfig struct {
 
 	// TargetContractsBalances holds the amount of wei that should be sent during deployment for one or more contracts in
 	// TargetContracts
-	TargetContractsBalances []*big.Int `json:"targetContractsBalances"`
+	TargetContractsBalances []ContractBalance `json:"targetContractsBalances"`
```
anishnaik, line 77:
```suggestion
	TargetContractsBalances []*ContractBalance `json:"targetContractsBalances"`
```

anishnaik: This may lead to additional code changes but I would recommend doing this.

fuzzing/config/config.go
```diff
@@ -112,11 +109,53 @@ type FuzzingConfig struct {
 	TestChainConfig config.TestChainConfig `json:"chainConfig"`
 }
 
-// fuzzingConfigMarshaling is a structure that overrides field types during JSON marshaling. It allows FuzzingConfig to
-// have its custom marshaling methods auto-generated and will handle type conversions for serialization purposes.
-// For example, this enables serialization of big.Int but specifying a different field type to control serialization.
-type fuzzingConfigMarshaling struct {
-	TargetContractsBalances []*hexutil.Big
+type ContractBalance struct {
```
anishnaik, line 112: Please make sure to add a comment for what `ContractBalance` represents above the struct definition.

fuzzing/config/config.go
```diff
@@ -112,11 +109,53 @@ type FuzzingConfig struct {
...
+		}
+	}
+
+	// Base-10 string handling
+	if _, ok := cb.Int.SetString(s, 10); !ok {
+		return fmt.Errorf("invalid base-10 string provided while unmarshaling contract balance: %s", s)
+	}
+	return nil
+}
+
```
anishnaik, line 156:
```suggestion
// MarshalJSON marshals a ContractBalance to JSON.
```

fuzzing/config/config.go
```diff
@@ -112,11 +109,53 @@ type FuzzingConfig struct {
...
+func (cb ContractBalance) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("\"%s\"", cb.Int.String())), nil
```
anishnaik, line 158:
```suggestion
	return []byte(cb.Int.String()), nil
```

anishnaik (changes requested):
Okay one more round of final nitpicks and review :)

0xZRA:
> Okay one more round of final nitpicks and review :)

No problem, will do

0xZRA:
@anishnaik Thank you for your comments, I hope I addressed latest round of feedback.

anishnaik:
Hey @0xZRA looks great! Can you run `npm prettier . --write` in the `docs/` folder? It will help the linter pass in the CI. PS you will probably have to download prettier which you can do via `npm -g install prettier`.

anishnaik (approved):

0xZRA:
@anishnaik just pushed the change after running `prettier`

anishnaik:
@0xZRA, it seems like `docs/theme/highlight.js` has been added to this PR which we don't want. Can you remove it from this PR?

anishnaik:
@0xZRA, I think there was a misunderstanding on my part around `highlight.js` (my mistake). What I meant was to revert the changes that `prettier` made to that `.js` file. But honestly it doesn't matter. Can you add the `.js` file back to this PR?

0xZRA:
> @0xZRA, I think there was a misunderstanding on my part around `highlight.js` (my mistake). What I meant was to revert the changes that `prettier` made to that `.js` file. But honestly it doesn't matter. Can you add the `.js` file back to this PR?

I feel like all these things could be caught and understood by me better before pushing to origin ;( Thanks for your patience

anishnaik (approved):

0xZRA:
@anishnaik I'm able to reproduce the test(s) failure locally but I'm not sure why sequence of `71` and `7` is triggering violated state or rather understand under which condition the failure became possible.
```
1) InheritedFirstContract.setY(uint256)(71) (block=4, time=165981, gas=12500000, gasprice=1, value=0, sender=0x30000)
2) InheritedFirstContract.setX(uint256)(7) (block=24092, time=602323, gas=12500000, gasprice=1, value=0, sender=0x10000)
[Execution Trace]
 => [call] InheritedFirstContract.setX(uint256)(7) (addr=0xA647ff3c36cFab592509E13860ab8c4F28781a66, value=0, sender=0x10000)
         => [return ()]

[Property Test Execution Trace]
[Execution Trace]
 => [call] InheritedFirstContract.property_never_specific_values()() (addr=0xA647ff3c36cFab592509E13860ab8c4F28781a66, value=0, sender=0x10000)
         => [return (false)]

⇾ Test summary: 1 test(s) passed, 1 test(s) failed
```

anishnaik:
Hey @0xZRA, honestly that specific test has been failing more frequently these days, I am unsure why. But luckily doesn't have anything to do with this PR. Going to merge now, thank you so much for the help and great job!

Issue #218, opened 2023-08-30
Code quality/project standards alignment


woodruffw:
This is a meta-issue, intended to track sub-issues that will align Medusa's codebase, development practices, etc. with the rest of our public tooling.

- [ ] Align `README.md` with our doc practices
- [ ] Streamline our installation steps
    - [ ] Move developer-specific installation instructions to their own `CONTRIBUTING.md` (this file already exists, but doesn't explain how to get a development environment stood up)
    - [ ] Provide a "single-step" install flow so that users don't have to manually download binaries from GitHub Releases
- [ ] Add some kind of `QUICKSTART.md` or similar guide

(This list is not intended to be exhaustive, and will be updated over time!)

PR #584, opened 2025-02-20
medusa-geth module path refactor

### LINKED ISSUE #205: One-line installation for medusa
Due to issues with forking and `go install` (https://github.com/golang/go/issues/44840), we need to identify a way to have one-line installation for all major platforms (outside of downloading the binaries directly).

We might have to pivot away from `go install` because there does not seem to be any momentum from the Golang team to loosen the restriction on the `replace` directive. More research into alternate solutions is required.

A decent solution could be to create a toolchain manager, similar to `foundryup`, that will manage installation and upgrades of medusa.

Xenomega:
This PR closes #205 (and addresses a bullet point in #218).

A primer on the current/previous state: 
- `go.mod` allows you to specify a `replaces` directive to redirect one package path to another. 
- `medusa` used this to redirect `github.com/ethereum/go-ethereum` paths to `github.com/crytic/medusa-geth`.
- This way, `medusa-geth` updates just involved forking and applying our patches on top, then re-linking medusa to the new version.
- However, the `replaces` directive only works on the top-level projects, so any attempt to use `medusa` as a dependency will not include the `replaces` directive and will break `medusa`'s ability to be imported. This same problem is also why `go install` will not work (as `replaces` is not respected).
- You should actually be doing a replace-all of the original module path with any forks then, it seems. Google discussed adding a new feature to better handle this in the future, but that has yet to be seen.

TLDR: We were prone to https://github.com/golang/go/issues/44840 .

This new PR updates this by:
- Refactoring all import/module paths in a new branch of `medusa-geth`.
- Refactor all import paths for `medusa-geth` in the `medusa` project.
- Replace the `replaces` directive with a proper import to `github.com/crytic/medusa-geth`.

The implications are:
- Currently open PRs may just need some touch ups/fixes on import paths if they added any new files.
- Supporting future forks of `go-ethereum` will now have two changes to their process: (1) a refactor commit doing a replace-all of any instances of `github.com/ethereum/go-ethereum` to `github.com/crytic/medusa-geth` should be done _after our patch set is applied and all is finalized_, (2) the patch set we carry forward should not include the _final refactoring commit (from step 1)_, as the imported files will likely change all the time.

Outstanding TODOs on my end before I ask for a PR review:
- [x] Test the medusa-geth patch applying cycle with the new process
- [x] Update the CONTRIBUTING.md in `medusa-geth` to reflect this.
- [ ] Verify no wiki/documentation should be updated elsewhere.

Currently, you can test this with `go install -v github.com/crytic/medusa@refactormod`. Once it is merged, `go install -v github.com/crytic/medusa@latest` should work


Xenomega:
This PR is now ready for review. It should be merged alongside https://github.com/crytic/medusa-geth/pull/2 (updated CONTRIBUTING.md in `medusa-geth` to reflect this new flow).

Keep in mind, this PR will be a bit painful in that it will change all module import paths, so existing PRs will probably have to replace any new/changed import paths to point at the new module path.

anishnaik (approved):
bing bong

PR #589, opened 2025-03-07
fix: return nil instead of error directly

laterlaugh:
Since we have already checked err before and returned when err != nil, err must be nil here.


Xenomega (approved):

Xenomega:
Good call, that pattern was not ideal. Thanks!

PR #594, opened 2025-03-13
Basic startPrank/stopPrank cheat code support

### LINKED ISSUE #446: Implement `startPrank` and `endPrank`

Xenomega:
This PR closes #446 . It introduces the basic variant of `startPrank` and `stopPrank` cheat codes (`msg.sender` support only)

Expected behavior:
- `startPrank` does pranking only one call level deeper than it was initiated from. Thus, the scope `startPrank` is called in is not affected, only direct sub-calls.
- Multiple `stopPrank` can be called without a `startPrank`, with no issue/revert.
- Multiple `startPrank` can be called, overriding the previous `startPrank` (it does not act like a stack).
  - Foundry will error if a previous `startPrank` was not applied at least once. This doesn't seem important to us to enforce(?)
- A `startPrank` in a reverted scope should inherently have no effect (not be pranking anymore).

Outstanding TODOs:
- [ ] Update the cheat code wiki page.
- [x] Open up issue to track adding support for `tx.origin` and `delegateCall` overloads in a later PR. The TX origin is exposed from `evm.Origin`, but we need an `evm` handle. And the `delegateCall` flag requires us to poke `scopeContext.Contract.self`, which is currently unexposed in `medusa-geth`. Relevant TODO comments are left as placeholders.


anishnaik (approved):
bing bong

PR #601, opened 2025-03-22
Add Configurable Verbosity Levels for Execution Traces

priyankabose:
Summary

This PR introduces three distinct verbosity levels for execution traces in Medusa, giving users more control over the level of detail shown in trace outputs. This is especially beneficial for complex systems with multiple call frames, where the current trace output can be overwhelming.

Changes

  - Added an enumerated VerbosityLevel type with three levels:
    - Verbose (0): Shows only top-level transactions in execution traces. Only events in the top-level call frame and return data are handled.
    - VeryVerbose (1): The current behavior (now default). Shows full detail for the current transaction.
    - VeryVeryVerbose (2): Shows full trace details for all call sequence elements, not just the last one (traceAll).
  - Added comprehensive test contracts and tests to verify the behavior of each verbosity level

### DIFF
--- a/cmd/fuzz_flags.go
+++ b/cmd/fuzz_flags.go
@@ -236,11 +236,23 @@ func updateProjectConfigWithFuzzFlags(cmd *cobra.Command, projectConfig *config.
 
 	// Update the verbosity levels
 	if cmd.Flags().Changed("verbosity") || cmd.Flags().Changed("v") {
-		verbosity, err := cmd.Flags().GetCount("verbosity")
+		verbosityCount, err := cmd.Flags().GetCount("verbosity")
 		if err != nil {
 			return err
 		}
-		projectConfig.Fuzzing.Testing.Verbosity = verbosity - 1
+
+		// Map verbosity count to VerbosityLevel enum
+		// -v = Verbose (0)
+		// -vv = VeryVerbose (1)
+		// -vvv = VeryVeryVerbose (2)
+		switch {
+		case verbosityCount == 1:
+			projectConfig.Fuzzing.Testing.Verbosity = config.Verbose
+		case verbosityCount == 2:
+			projectConfig.Fuzzing.Testing.Verbosity = config.VeryVerbose
+		case verbosityCount >= 3:
+			projectConfig.Fuzzing.Testing.Verbosity = config.VeryVeryVerbose
+		}
 	}
 
 	return nil
--- a/cmd/fuzz_flags.go
+++ b/cmd/fuzz_flags.go
@@ -77,8 +77,8 @@ func addFuzzFlags() error {
 	// RPC block
 	fuzzCmd.Flags().Uint64("rpc-block", 0, "block number to use when fetching contracts over RPC")
 
-	// Verbosity levels
-	fuzzCmd.Flags().CountP("verbosity", "v", "use verbose output")
+	// Verbosity levels (-v, -vv, -vvv)
+	fuzzCmd.Flags().CountP("verbosity", "v", "increase verbosity level (can be used multiple times: -v, -vv, -vvv)")
 	return nil
 }
 
--- a/cmd/fuzz_flags.go
+++ b/cmd/fuzz_flags.go
@@ -56,10 +56,6 @@ func addFuzzFlags() error {
 	fuzzCmd.Flags().String("deployer", "",
 		"account address used to deploy contracts")
 
-	// Trace all
-	fuzzCmd.Flags().Bool("trace-all", false,
-		fmt.Sprintf("print the execution trace for every element in a shrunken call sequence instead of only the last element (unless a config file is provided, default is %t)", defaultConfig.Fuzzing.Testing.TraceAll))
-
 	// Logging color
 	fuzzCmd.Flags().Bool("no-color", false, "disables colored terminal output")
 
@@ -80,6 +76,9 @@ func addFuzzFlags() error {
 
 	// RPC block
 	fuzzCmd.Flags().Uint64("rpc-block", 0, "block number to use when fetching contracts over RPC")
+
+	// Verbosity levels
+	fuzzCmd.Flags().CountP("verbosity", "v", "use verbose output")
 	return nil
 }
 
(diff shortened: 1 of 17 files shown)


cmd/fuzz_flags.go
```diff
@@ -243,5 +234,26 @@ func updateProjectConfigWithFuzzFlags(cmd *cobra.Command, projectConfig *config.
 		}
 	}
 
+	// Update the verbosity levels
+	if cmd.Flags().Changed("verbosity") || cmd.Flags().Changed("v") {
+		verbosityCount, err := cmd.Flags().GetCount("verbosity")
```
anishnaik, line 239: Will `GetCount` for "verbosity" also handle the counts for "v" as well?

fuzzing/executiontracer/execution_trace.go
```diff
@@ -34,14 +35,17 @@ type ExecutionTrace struct {
 
 	// labels is a mapping that maps an address to its string representation for cleaner execution traces
 	labels map[common.Address]string
+
```
anishnaik, line 38:
```suggestion
// verbosity describes the verbosity levels of the execution trace
```

fuzzing/executiontracer/execution_tracer.go
```diff
@@ -69,14 +70,17 @@ type ExecutionTracer struct {
 
 	// nativeTracer is the underlying tracer interface that the execution tracer follows
 	nativeTracer *chain.TestChainTracer
+
```
anishnaik, line 73:
```suggestion
// verbosity describes the verbosity level that will be used for the execution trace
```

fuzzing/fuzzer.go
```diff
@@ -556,7 +556,7 @@ func chainSetupFromCompilations(fuzzer *Fuzzer, testChain *chain.TestChain) (*ex
 					if err != nil {
 						return nil, fmt.Errorf("failed to reset to genesis block: %v", err)
 					} else {
-						_, err = calls.ExecuteCallSequenceWithExecutionTracer(testChain, fuzzer.contractDefinitions, []*calls.CallSequenceElement{cse}, true)
+						_, err = calls.ExecuteCallSequenceWithExecutionTracer(testChain, fuzzer.contractDefinitions, []*calls.CallSequenceElement{cse}, 2)
```
anishnaik, line 559: I think we should pass `config.Verbose` here since the trace will only have 1 top-level message.

fuzzing/testdata/contracts/execution_tracing/verbosity_levels.sol
```diff
@@ -0,0 +1,225 @@
...
+     */
+    function delistItem(uint256 itemId) public {
+        require(listings[itemId].active, "Item not listed");
+        require(listings[itemId].seller == msg.sender, "Not the seller");
+
+        listings[itemId].active = false;
+
+        emit ItemDelisted(itemId);
+    }
+}
```
anishnaik, line 225: Add newline at the end of this test contract

docs/src/project_configuration/testing_config.md
```diff
@@ -62,12 +62,15 @@ contract MyContract {
...
-- **Default**: `false`
+### `verbosity`:
+
+- **Type**: Enum (VerbosityLevel)
+- **Description**: Controls the level of detail in execution traces:
+  - **Verbose (0)**: Only shows top-level transactions in the execution trace. Only events in the top-level call frame and return data are included.
+  - **VeryVerbose (1)**: Shows nested calls with standard detail (default behavior).
+  - **VeryVeryVerbose (2)**: Shows all call sequence elements with maximum detail, attaching traces to every call in the sequence.
+- **CLI flags**: Use `-v`, `-vv`, or `-vvv` to set verbosity levels 0, 1, or 2 respectively.
+- **Default**: `VeryVerbose (1)`
```
anishnaik, line 73: Can you also update the documentation for the CLI? See `docs/src/cli/fuzz.md`.

cmd/fuzz_flags.go
anishnaik: More generally, can you make sure that the feature is working directly through the CLI and not only using unit tests?

fuzzing/testdata/contracts/execution_tracing/verbosity_levels.sol
```diff
@@ -0,0 +1,225 @@
...
+ * @title Verbosity Levels Test Contract
+ * @dev This file contains contracts specifically designed to test Medusa's verbosity levels:
+ *
+ * - Verbose (Level 0): Only shows top-level calls, hides nested calls
+ * - VeryVerbose (Level 1): Shows nested calls with event and return data (default)
+ * - VeryVeryVerbose (Level 2): Shows all call sequence elements with event and return data
+ *
+ * The test creates a complex call structure with nested function calls across multiple contracts:
+ * 1. TestMarketplace.buyItem calls an internal function _processPayment
+ * 2. _processPayment calls external functions on the TestToken contract:
```
anishnaik, line 14: Although this unit test is good, it is rather complicated for what we are testing. A simpler `A` calls `B` calls `C` structure where `C.f()` throws an `assert(false)` should be sufficient to test this feature. We want unit tests to be performant (since we run it in the CI) while still testing the core feature. Thus, doing the A->B->C where you throw in some events and return values scattered here and there should be sufficient. You can probably re-use one of the other execution trace test cases for this one.

anishnaik (changes requested):
The core functionality looks great! My only gripe is the complexity of the smart contract test case. A simpler test is always better as long as it covers the same code paths as the test case you have written (which I believe it can).

Additionally, can you make sure that the feature works as expected through the CLI? Thus, build the binary and run `medusa fuzz -v`, `medusa fuzz -vv`, and `medusa fuzz -vvv` on the same piece of test code. You can run these commands on the same contract that you wrote the unit test for.

fuzzing/fuzzer.go
anishnaik: Ahh right good point, okay let's update it to `config.VeryVeryVerbose` instead of `2`.

priyankabose: Updated!

cmd/fuzz_flags.go
priyankabose: Yes, tested using CLI first and then added the unit test.

docs/src/project_configuration/testing_config.md
priyankabose: Added.

fuzzing/testdata/contracts/execution_tracing/verbosity_levels.sol
priyankabose: Added a more simplified the test case. We still need a bit of a state to distinguish between VeryVerbose and VeryVeryVerbose mode.

fuzzing/executiontracer/execution_trace.go
priyankabose: Added the change

fuzzing/executiontracer/execution_tracer.go
priyankabose: Added

anishnaik (approved):

PR #609, opened 2025-03-28
Fix `go install` installation

elopez:
```sh
$ go install github.com/crytic/medusa@latest
go: github.com/crytic/medusa@latest (in github.com/crytic/medusa@v1.2.0):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

This is caused by a leftover replace in `go.mod`. Medusa no longer references `github.com/ethereum/go-ethereum` so it should be ok to remove.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -8,7 +8,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const version = "1.2.0"
+const version = "1.2.1"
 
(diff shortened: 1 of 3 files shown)


anishnaik:
FFFFFFFFFFFF how did I miss this

anishnaik (approved):

Issue #611, opened 2025-04-17
Allow Medusa to run init function (ex: setUp) once per fuzz run


montyly:
A common pattern for large codebase is to setup the system through the `setUp()` function (from [foundry](https://book.getfoundry.sh/forge/writing-tests#writing-tests)).

It would be great if the user could specify an "init function" (which would be `setUp()` by default). 

Given that medusa can take a list of contracts to deploy in `targetContracts`, this probably requires a dictionary in the config to configure the init function per contract

Wdyt?

anishnaik:
Joss! We meet again :) 

Yes, I think we should have this - I know we have a rogue PR somewhere with the building blocks for this. This will be low priority since we have the medusa stylus grant working ongoing.

anishnaik:
@0xZRA do you want to tackle this issue instead of working on the documentation?

0xZRA:
@anishnaik sure, I can work on this!

anishnaik:
I guess before we kick this off, @montyly what would be the benefit of using `setUp` or a generic `init` function over just using the constructor?

montyly:
If you need a callback in the initialization, then you can't do it in the constructor. 

Ex:
- A's `setUp` function creates `B`, and call `B.some_function`
-  `B.some_function` calls `A.some_other_function`

Which will not work if you use A's constructor instead of `setUp` (given that there is no code deployed yet)

PR #618, opened 2025-04-25
[WIP]: Allow Medusa to run init function (ex: setUp) once per fuzz run

0xZRA:
Closes https://github.com/crytic/medusa/issues/611


0xZRA:
> @0xZRA appreciate the quick turnaround! I need a few days before I can get to reviewing it just FYI :)

Absolutely! And thanks for the opportunity ;)

montyly:
My bad, I never noticed the progression on this PR. This is great thanks @0xZRA. I tested it on a larger codebase and it works

In term of usability, I am wondering it would make sense to add a flag to enable `setUp()` to be called by default on all the contract that implements it?

0xZRA:
@montyly Thanks for your feedback! I can work on adding the flag, it may take mew few days though

rs-ifflabs:
> My bad, I never noticed the progression on this PR. This is great thanks @0xZRA. I tested it on a larger codebase and it works
> 
> In term of usability, I am wondering it would make sense to add a flag to enable `setUp()` to be called by default on all the contract that implements it?

@montyly Let me know if the latest commit delivers the intended improvement to devx. My assumption is that the default configuration values will still be processed downstream to address all contracts that implement it. 
I couldn't find the test that covers flags so maybe it's a good opportunity to add it?

0xZRA:
@montyly Let me know if the latest commit delivers the intended improvement to devx. My assumption is that the default configuration values will still be processed downstream to address all contracts that implement it.
I couldn't find the test that covers flags so maybe it's a good opportunity to add it?

0xZRA:
@montyly let me know if the latest approach works.
I also added the `UseInitFunctions` to control whether init functions (via `use-init-fns` or `enable-foundry-setup`) should be processed at all.
Since you mentioned that the main functionality seemed to be working as expected, the initialization logic hasn't changed, it's just now controlled by the feature flag.

My main concern is the number of changes to `fuzzing/fuzzer.go` since I worked on this task and it might take me a while to get up to speed on those changes before I can resolve the conflicts.

anishnaik:
Thanks for giving this a review @montyly! @0xZRA apologies for the massive delay. Will get this merged in the next few weeks.

anishnaik:
@codex review

0xZRA:
> Sounds good @0xZRA. I did a brief review of this this week and I think that this PR is more complicated than it should be. Although I am a fan of providing maximum flexibility to users, I don't see a lot of benefit in providing the ability to have input arguments (you can already do that with constructors) and having unique names for the `setUp` function. I would rather merge a PR that simply allows you to call the `setUp` function so we have more compatibility with Foundry. Ideally, when we are parsing the target contracts, we search for the `setUp` ABI function and ensure that those functions are called during contract deployment. Does that make sense?
> 
> I apologize for not telling you this sooner but in general I don't want to overcomplicate the system logic unless I see an obvious use case for this.

@anishnaik I follow and understand your reasoning. I can definitely simplify it and I could realistically get to it within a week, but also if it's holding up any of your work feel free to take it over. I completely understand and wouldn't want to hold anyone up

PR #625, opened 2025-04-29
Prune unnecessary transaction sequences from corpus

samalws-tob:
This PR adds a job which runs once a minute that prunes unnecessary txn sequences from the corpus. It removes sequences that no longer contribute any new coverage compared to the rest. This can happen when, for example, sequence A adds new coverage, sequence B builds upon sequence A and adds strictly more coverage, making sequence A now unnecessary since it doesn't "contribute anything" to the corpus.
The job works by:
- Shuffle the corpus
- Initialize a blank coverage map
- Run through the corpus one-by-one. For each sequence, run the sequence and add coverage to the coverage map. If there is no new coverage, we can remove this sequence from the corpus

The job takes about 5-10 seconds, during which time no fuzzing at all can occur

This seems to be pretty effective at reducing corpus size. In a small test I've been running, it's been removing about 100 entries per minute during the initial phase of fuzzing where the corpus quickly grows to 1000 entries (first 5 minutes) and has now slowed down to removing about 50 entries per minute (15 minutes in, corpus size 1400). Corpus growth is slow at this phase because of pruning; the corpus would probably have 2000-2500 entries by now (rather than 1400) if not for pruning.


fuzzing/corpus_pruner.go
```diff
@@ -0,0 +1,98 @@
...
+func (cp *CorpusPruner) pruneCorpus() error {
+	start := time.Now() // We'll track how long pruning takes
+	n, err := cp.fuzzer.corpus.PruneSequences(cp.fuzzer.ctx, cp.chain)
+	// PruneSequences takes a while, so ctx could've finished in the meantime.
+	// If it did, we skip the log message.
+	if err != nil || utils.CheckContextDone(cp.fuzzer.ctx) {
+		return err
+	}
+	cp.totalCorpusPruned += n
+	cp.fuzzer.logger.Info(fmt.Sprintf("Pruned %d values in %v. Total pruned this run: %d", n, time.Since(start), cp.totalCorpusPruned))
```
anishnaik, line 50:
```suggestion
	cp.fuzzer.logger.Info(fmt.Sprintf("Pruned %d corpus items in %v seconds. Total pruned this run: %d", n, time.Since(start), cp.totalCorpusPruned))
```

fuzzing/corpus_pruner.go
```diff
@@ -0,0 +1,98 @@
...
+	for {
+		select {
+		case <-cp.fuzzer.ctx.Done():
+			return
+		case <-ticker.C:
+			err := cp.pruneCorpus()
+			if err != nil {
+				cp.fuzzer.logger.Error("Corpus pruner encountered an error", err)
+				cp.fuzzer.Terminate() // We kill the whole thing if the pruner errors
+				return
```
anishnaik, line 69: Personally, I don't think we need to shut down the fuzzer if there is an error. Since the pruner is not essential to the core fuzzing loop, I think we can log the error and move on.

fuzzing/corpus_pruner.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// CorpusPruner is a job that runs every `PruneFrequency` minutes.
+// It removes unnecessary items from the corpus by calling `Corpus.PruneSequences`.
+type CorpusPruner struct {
+	// enabled determines if the pruner is enabled
+	enabled bool
+
+	// fuzzer describes the Fuzzer instance which this pruner belongs to
+	fuzzer *Fuzzer
```
anishnaik, line 19: This is a nitpick but I don't think we need to have access to the fuzzer here. Instead, I think access only to the corpus should be sufficient? Especially if we don't call `fuzzer.Terminate` (see my other comment) and allow the corpus pruner to have its own sub-logger.

Compositionally I think this is better because it's called the "corpus pruner" so it shouldn't really have a dependency on the fuzzer.

fuzzing/corpus/corpus.go
```diff
@@ -543,3 +553,77 @@ func (c *Corpus) Flush() error {
...
+		seqs[i] = seqCloned
+	}
+	c.callSequencesLock.Unlock()
+	// We don't need to lock during the next part as long as the ordering of Choices doesn't change.
+	// New items could get added in the meantime, but older items won't be touched.
+
+	toRemove := map[int]bool{}
+
+	// Iterate seqs in a random order
+	for _, i := range rand.Perm(len(seqs)) { // TODO should we use a "random provider"?
```
anishnaik, line 589: Eh, I don't think it matters if we use a random provider or not.

fuzzing/corpus/corpus.go
```diff
@@ -543,3 +553,77 @@ func (c *Corpus) Flush() error {
...
+		if utils.CheckContextDone(ctx) {
+			return 0, nil
```
anishnaik, line 591: It doesn't matter too much I guess but shouldn't we return `return toRemove, nil`?

fuzzing/corpus_pruner.go
samalws-tob: The %v already puts a unit afterwards. Preferring %v because it displays it nicely if it's milliseconds, like '800ms' rather than '.8s'

fuzzing/corpus_pruner.go
samalws-tob: We'd need to pass corpus, logger, context, and PruneFrequency, which imo is fine

samalws-tob: If we do that we could probably also move the file into the corpus subdirectory, since I only had it in fuzzer/ so it could access the fuzzer struct

fuzzing/corpus_pruner.go
samalws-tob: I was thinking it might invalidate the state during an error but then worst case a worker would error out and we'd exit anyway

samalws-tob: removed terminate call

fuzzing/corpus_pruner.go
samalws-tob: Changed 'values' -> 'corpus items'

anishnaik (approved):

PR #630, opened 2025-05-09
Add external library support

priyankabose:
``Overview``

This PR adds support for external libraries in Medusa, allowing for proper deployment, linking, and testing of contracts that depend on external Solidity libraries.

``Key Changes``

  - Added infrastructure to identify and map library dependencies between contracts
  - Implemented topological sorting algorithm for determining correct deployment order
  - Created bytecode linking functionality to replace library placeholders with deployed addresses
  - Updated test chain deployment process to respect library dependency order

  ``Testing``

  - Added test contracts in `fuzzing/testdata/contracts/external_library/` demonstrating various library dependency scenarios
  - Test contracts include multi-level dependencies (libraries depending on other libraries)
  - Tests validate correct library deployment order and proper function execution through dependencies

### DIFF
--- a/compilation/platforms/crytic_compile.go
+++ b/compilation/platforms/crytic_compile.go
@@ -1,7 +1,6 @@
 package platforms
 
 import (
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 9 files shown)


anishnaik (approved):
bing bong

PR #660, opened 2025-08-04
Add flag: returnBool for property testing

montyly:
`returnBool` set to false will make medusa accept properties that don't return a boolean.

This is meant to mimic foundry's behavior, and ease the compatibility between the two tools. See https://getfoundry.sh/forge/advanced-testing/invariant-testing/#conditional-invariants


montyly:
@elopez the way I have seen people using it in foundry setup is basically that the `invariant_` functions should not revert.

It would be the same as having `require(return_variable)`

PR #716, opened 2025-11-24
ci: check vendorHash and nix build on PRs

elopez:
This should alert folks with a comment when they update go.mod / go.sum but forget to refresh flake.nix to match.


elopez:
@codex review

PR #717, opened 2025-11-24
Debug full solidity projects

anishnaik:
There was no way before this PR to step through and debug a medusa error that occurs in a codebase. Our unit test framework was limited to a specific solidity file but now we can point to a random filepath on the local filesystem and run medusa on it. 

Note that this is pretty barebones. For example, if `medusa.json` is not in the base project path then it will throw a fit. We can always extend this if necessary but should be good enough for now.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,15 +2,19 @@
   "nodes": {
     "crytic": {
       "inputs": {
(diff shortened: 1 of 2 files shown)


anishnaik:
@codex review

Issue #793, opened 2026-02-06
Add thorough documentation for agents


anishnaik:
Inspired by this article: https://vercel.com/blog/agents-md-outperforms-skills-in-our-agent-evals

We should maintain robust documentation for every component of medusa so that future bug fixes and features can be implemented rapidly with the necessary context.

PR #832, opened 2026-07-06
docs: add agent-focused OpenWiki documentation and fix stale AGENTS.md claims

nisedo:
## Summary

I used [OpenWiki](https://www.langchain.com/blog/introducing-openwiki-an-open-source-agent-for-repo-documentation) with Anthropic's Claude models (Fable 5 / Opus 4.8) to create thorough, agent-focused documentation for medusa and to audit our existing agent docs against the source code.

References #793.

Two commits:

1. **`openwiki/` — a source-grounded documentation layer for LLM agents** (plus a pointer section in `CLAUDE.md`). It's deliberately a *map/synthesis* layer: it explains how the codebase is organized and where to change things, linking to the mdBook for end-user how-tos instead of duplicating them. Pages:
   - `quickstart.md` — what medusa is, repo layout, quick-run commands
   - `architecture/overview.md`, `architecture/fuzzing-engine.md`, `architecture/chain-and-state.md`
   - `workflows/fuzzing-lifecycle.md` — end-to-end campaign flow
   - `testing.md`, `operations.md` (CLI + `ProjectConfig` schema with defaults)
   - `reference/source-map.md` — "I want to change X, where do I look?" index
2. **`AGENTS.md` accuracy fixes.** Every factual claim in the generated wiki *and* in the existing `AGENTS.md` was verified claim-by-claim against the code (~450 claims, with file:line evidence). The wiki fixes are already folded into the first commit; this commit fixes the stale/incorrect items found in `AGENTS.md`, including:
   - `cmd/` command list was missing `corpus clean`
   - `go run . --config path/to/config.yaml` → `go run . fuzz --config path/to/medusa.json` (`--config` is a `fuzz` flag; configs are JSON)
   - `OptimizationTestCaseProvider` maximizes an `int256` test-function return value, not "gas usage"
   - corpus is persisted only when `corpusDirectory` is set (default is in-memory)
   - deployment happens once on a base chain that workers clone (not per worker); test hooks run per call, not per sequence
   - property tests: `view`/`pure` is convention, not enforced by the classifier
   - added the missing `docs-check` pre-commit hook and `slither` config section; noted that CI enforces `prettier` while the local hook uses `dprint`

The same audit also covered `README.md` (clean) and the mdBook under `docs/src` — I'll submit the mdBook fixes (45 findings, e.g. non-existent `--assertion-mode`/`--optimization-mode` flags and a `coverageReports` → `coverageFormats` key error) as a separate PR to keep this one reviewable.

## Test plan

- `python3 scripts/check_docs.py` passes
- `prettier --check` / `dprint check` unaffected (new files are markdown-only additions)
- All wiki links resolve to existing files; all cited symbols/defaults verified against source at `66bb59d`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,27 +23,27 @@ Medusa is a cross-platform smart contract fuzzer written in Go, built on go-ethe
 
 ### Lint & Format
 
-- `goimports -w .` - Format imports and code (groups stdlib, external, local)
+- `goimports -w .` - Format imports and code (groups stdlib vs. external; no `-local` grouping is configured)
 - `go fmt ./...` - Format Go code (required before commits)
 - `golangci-lint run --timeout 5m` - Run comprehensive lint checks (mirrors CI)
-- `dprint fmt` - Format markdown, YAML, and JSON files
+- `dprint fmt` - Format markdown, YAML, and JSON files (note: CI enforces `prettier --check` on `**.json`/`**/*.md`/`**/*.yml`, which can disagree with dprint — run prettier before pushing)
 - `actionlint` - Lint GitHub Actions workflow files
 
 ### Run
 
-- `go run . --config path/to/config.yaml` - Compile and run the fuzzer
+- `go run . fuzz --config path/to/medusa.json` - Compile and run the fuzzer (`--config` is a flag of the `fuzz` subcommand; configs are JSON)
 - `nix develop` - Enter the pinned Nix development shell with all dependencies
 
 ### Git Hooks
 
 - `prek install` - Install pre-commit hooks (run once after cloning)
-- `prek run` - Manually run all pre-commit checks (use within `nix develop` shell)
+- `prek run` - Manually run all pre-commit checks (the hook tools are provided by the `nix develop` shell; `prek` itself is installed separately)
 
 ## Architecture & Code Structure
 
 ### Module Responsibilities
 
-- **`cmd/`** - CLI entry point using Cobra framework. Defines `fuzz`, `init`, and `completion` commands.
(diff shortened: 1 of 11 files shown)


nisedo:
### Note on the failing `lint` check — pre-existing CI breakage, unrelated to this PR

This PR is markdown-only, but the `lint` job fails with Go `typecheck` errors. Root cause analysis:

**The golangci-lint binary can no longer typecheck the Go stdlib installed on current runners.** From the failing job log:

```
go version go1.26.4 linux/amd64
##[error].../go/1.26.4/x64/src/vendor/golang.org/x/crypto/chacha20poly1305/fips140only_go1.26.go:7:9:
  file requires newer Go version go1.26 (application built with go1.24) (typecheck)
```

- The lint job uses `setup-go` with `go-version: "^1.22"` (open-ended range) + pinned `golangci-lint v1.64.8` (built with Go 1.24).
- On master's last green run (May 14, [76043760250](https://github.com/crytic/medusa/actions/runs/28779454222)) `^1.22` resolved to **Go 1.25.9** → passed. Today it resolves to **Go 1.26.4** (runner image update) → golangci-lint v1.64.8 cannot load the 1.26 stdlib, and module loading fails.
- All the downstream `undefined: Message` / `tracer.OnTxStart undefined` errors are fallout: `chain/vendored/apply_transaction.go` dot-imports `medusa-geth/core`, so when the linter fails to load modules, every imported symbol reads as undefined. The same tree compiles and passes `go test` on ubuntu/macos/windows in this very run.
- Reproduced deterministically on a re-run; any branch (including master) would fail this job today.

**Suggested fix** (separate PR): pin the lint job's `go-version` to `"1.24"` (matches `go.mod` and the pinned linter), or bump golangci-lint to the v2 line.

PR #835, opened 2026-08-29
refactor: use typed event reflection and zero-size sets

zloglevel:
## Description

Use `reflect.TypeFor[T]()` where generic event types are known at compile time, and represent the address/string membership sets with `map[K]struct{}` instead of `map[K]any`. This makes the reflection intent explicit and avoids storing an interface value for each set member.

**Related Issues:** N/A

**Type of Change:**

- [x] Refactoring
- [x] Performance Improvement

## Testing

- `go test ./events`
- `go test ./fuzzing/valuegeneration`

## Additional Context

Membership behavior, cloned values, event keys, and public APIs are unchanged.

## Outstanding TODOs

None.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -10,7 +10,7 @@ type EventHandler[T any] func(T) error
 
 // EventType returns the event type given an EventHandler object
 func (e *EventHandler[T]) EventType() reflect.Type {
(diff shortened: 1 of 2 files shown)


zloglevel:
@Xenomega  Could you please take a look when you have a chance? Thanks!

